PR #4, opened 2021-05-21
Packaging

jonathangreen:
## Description

- Updates the documentation for the fork
- Updates the documentation to reflect [ADR-2](https://www.notion.so/lyrasis/Distribute-admin-UI-with-jsdelivr-417a5accc475480788e90fb74b9bbe18) 
  - Including adding development documentation
- Add license file to repo
- Modify the webpack configuration slightly to reduce filesize
- Be more selective about the files we put in our package

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,5 @@
 src/*
+tests/*
 lib/*
 typings/*
 *~
@@ -10,3 +11,8 @@ coverage
 
 # github
 .github
+
+.sass-lint.yml
+/*.json
+/webpack.*.js
+/.*
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -52,7 +52,11 @@ module.exports = {
       },
       {
         test: /\.(ttf|woff|eot|svg|png|woff2|gif|jpg)(\?[\s\S]+)?$/,
-        loader: 'url-loader?limit=100000'
+        loader: 'url-loader',
+        options: {
+            'limit': 10000,
+            'name': 'assets/[hash].[ext]'
+        }
       }
     ]
   },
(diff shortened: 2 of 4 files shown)


tdilauro (approved):
This looks good to me.

A couple things:
- I'm not strong on the front-end stuff, so it may be good to have Ray look at the `.npmignore` and webpack config changes.
- You didn't touch this, but I was wondering about the Nightwatch/Selenium stuff mentioned in the README. Is that really running somewhere?

jonathangreen:
@ray-lee these changes are very similar to the changes in https://github.com/ThePalaceProject/library-registry-admin/pull/3. Do you mind taking a look at the webpack config and `.npmignore` changes here as well?

.npmignore
```diff
@@ -10,3 +11,8 @@ coverage
 
 # github
 .github
+
+.sass-lint.yml
+/*.json
+/webpack.*.js
+/.*
```
ray-lee, line 18: This is fine, but at some point we may want to whitelist via the `files` property in `package.json`. https://docs.npmjs.com/cli/v7/configuring-npm/package-json#files

webpack.common.js
```diff
@@ -52,7 +52,11 @@ module.exports = {
       },
       {
         test: /\.(ttf|woff|eot|svg|png|woff2|gif|jpg)(\?[\s\S]+)?$/,
-        loader: 'url-loader?limit=100000'
+        loader: 'url-loader',
+        options: {
+            'limit': 10000,
+            'name': 'assets/[hash].[ext]'
+        }
```
ray-lee, line 59: Same comment as #3: We should check if files that are not inlined are able to be loaded by the app when the JS bundle is loaded from a CDN.

PR #8, opened 2021-07-12
Sync from Simplified PR #350 / commit 6f88c82

tdilauro:
## Description

Sync up with [Simplified circulation-web](https://github.com/NYPL-Simplified/circulation-web) v0.5.1+ ([PR #350](https://github.com/NYPL-Simplified/circulation-web/pull/350) / [commit 6f88c82](https://github.com/NYPL-Simplified/circulation-web/commit/6f88c82ff74e5f88ca2c8202ed862fc9e4d4ec5f)).

Contains:
- Some security fixes.
- More systemic fix for the footer layout issue.
- "Added a custom Mocha reporter so that a formatted list of suites and tests can be generated via the CLI."
- Clarifications in README.md.

I also had to make a few changes extra changes: 
- Address restrictions imposed by the `npx lint-staged` check that runs as a pre-commit hook.
- Dropped our PR template and accepted the upstream one to make future syncs easier. They are identical and result  in identical behavior. They were just in different places.

## Checklist:


ray-lee (approved):
Looks good to me! (I really only reviewed the last commit from Tim, on the assumption we want everything from upstream.)

PR #10, opened 2021-09-25
Merge from simplified

tdilauro:
## Description

Sync up with Simplified circulation-web v0.5.5+ (PR [#361](https://github.com/NYPL-Simplified/circulation-web/pull/361) / commit [c29255c](https://github.com/NYPL-Simplified/circulation-web/commit/c29255ce5441363df954b550f4d7b4ee4fbc78b9)).

Contains:
- Updated opds-web-client to 0.6.3 to fix a List Manager bug.
- Updated node-sass and sass-loader dependency versions to reduce the number of high risk vulnerabilities.
- Updated CustomListEditor so that the "Save this list" button is disabled if either title or entries are absent, and enabled if both are present.
- Updated the book cover editor so that the current cover URL pulls through not as a value in the input (as described in v0.5.3), but as a description beneath the input.
- Added a value attribute to the book cover URL's EditableInput in BookCoverEditor so that the existing URL pulls through for the user to start.
- Prepended all deprecated React lifecycle methods (componentWillMount and componentWillReceiveProps) with UNSAFE_.

## Checklist:


jonathangreen (approved):
@ray-lee is probably a better reviewer for this then me, but it looks okay to me.

ray-lee (approved):
The effectiveness of this review is questionable, because there is just a huge number of formatting changes and `let` → `const` conversions, and I ended up only skimming everything. But I didn't catch anything obviously bad.

tdilauro:
I'll take "not obviously bad" at this point. Thank you both!

PR #11, opened 2021-09-28
Avoid some columns collapsing when other columns contain long names/titles.

tdilauro:
## Description

Specifies minimum width for some CSS selectors to prevent columns collapsing when other columns contain long names/titles

## Motivation and Context

This is affecting the custom list manager for several of our libraries.

## How Has This Been Tested?

Manual testing by modifying values in the browser developer tools.

## Checklist:

- N/A I have updated the documentation accordingly.


ray-lee (approved):
Looks good to me!

PR #12, opened 2021-10-12
Add npm dev-server script to develop against a remote CM back-end.

ray-lee:
## Description

This adds a new npm script, `dev-server`, which runs a local development server that proxies requests to a specified CM back-end.

It also:
- Upgrades the version of webpack-cli used by the project.
- Changes the source map generation method in dev to the recommended (by webpack) `eval-source-map`.

## Motivation and Context

This allows development on circulation-admin to happen without the need to build a local circulation manager from source. It makes it easier for front-end developers to get started on the project, as they don't need to install a python ecosystem, elasticsearch, etc. It also allows data from "real" CMs to be used during development, without needing to dump and restore databases.

## How Has This Been Tested?

1. Run `npm run dev-server -- --env=backend=https://gorgon.tpp-qa.lyrasistechnology.org`
2. Open http://localhost:8080/admin in a browser
3. Log in with gorgon credentials. Verify that content from gorgon appears.
4. Edit some source code. The browser should automatically reload, and the expected changes should appear.

## Checklist:

### DIFF
--- a/webpack.dev-server.config.js
+++ b/webpack.dev-server.config.js
@@ -122,7 +122,7 @@ module.exports = (env) => {
   };
 
   /**
-   * Rewrites an HTML response. This changes jsdelivr CDN URLs in the page to point to the webapack
+   * Rewrites an HTML response. This changes jsdelivr CDN URLs in the page to point to the webpack
    * assets on the local server instead. This is a simple find-and-replace.
    *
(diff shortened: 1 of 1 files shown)


ray-lee:
I think some of the readme changes need to be there to make the new stuff make sense, and it's an improvement, even if it's still not correct. I'd rather leave it, and come back to fix all the problems.

PR #15, opened 2021-12-15
Update config tabs for librarians and library managers.

ray-lee:
## Description

1. On the System Configuration screen, remove the Analytics tab for librarians and library managers, and add the Admins tab.
2. Making the Admins tab available to librarians revealed some problems. Even though librarians (non-managers) don't have the ability to do any useful edits, the UI implied that they could. The Admins tab is now properly read-only for librarians, with the following changes:
   1. The Create button was present. Librarians are actually able to create users, but can't assign them to any roles in any libraries, so the users can't log in. The Create button is now not shown.
   2. Edit buttons were present for each user. This opened a form where all the inputs were disabled. The Edit buttons are now replaced with View buttons.
   3. Clicking the Edit (now View) button for a user opened a form with the title "Edit {username}", even though all the inputs were disabled. The title is now just the username.
   4. The form for a user had an enabled Submit button, even though all the inputs were disabled. The Submit button is now not shown.
3. Relabel the "library manager" role to "administrator", and "librarian" role to "user".

## Motivation and Context

This addresses issues raised in this Notion ticket (and its comments): https://www.notion.so/lyrasis/Remove-Analytics-and-add-Admins-to-Library-Manager-view-in-CM-5d0a2758a99d4b86881a9485d9be9a2b

From the ticket:
- On our old CMs, the Libraries and Admins tabs were shown to library managers. NYPL changed this to Libraries and Analytics at some point. We still want to allow library staff to view and edit their logins.
- "administrator" and "user" instead of "library manager" and "librarian" match how the two tiers are listed in the Palace Marketplace.

## How Has This Been Tested?

1. I logged in as a system admin, and verified that:
   - In the user dropdown, the first line reads "Logged in as a system admin".
   - In System Configuration, all tabs still appear.
   - On the Admins tab, I can still create, edit, and delete users.
   - On the edit user form, the "library manager" role is now referred to as "administrator", and the "librarian" role is now referred to as "user".
2. I logged in as a library manager, and verified that:
   - In a library catalog, in the user dropdown, the first line reads "Logged in as an administrator".
   - In System Configuration, the tabs shown are Libraries and Admins.
   - On the Admins tab, I can create and edit users.
3. I logged in as a librarian, and verified that:
   - In a library catalog, in the user dropdown, the first line reads "Logged in as a user".
   - In System Configuration, the tabs shown are Libraries and Admins.
   - The Admins tab is read-only, and the UI doesn't imply that I can edit anything.
      - There is no Create button.
      - The buttons next to users say "View", not "Edit".
      - After clicking on a user, the title doesn't say "Edit", all inputs are disabled, and the Submit button does not appear.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,12 @@
 ## Changelog
 
-### October 28, 2021
+### December 13, 2021
 
 #### Updated
 
 - Updated branding and styling for The Palace Project.
 - Updated node-sass and sass-loader dependency versions to reduce the number of high risk vulnerabilities.
+- Removed Analytics tab from System Configuration for librarians and library managers, and added Admins tab.
 
 ### v0.5.5
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,13 @@
 ## Changelog
 
-### December 13, 2021
+### December 14, 2021
 
 #### Updated
 
 - Updated branding and styling for The Palace Project.
 - Updated node-sass and sass-loader dependency versions to reduce the number of high risk vulnerabilities.
 - Removed Analytics tab from System Configuration for librarians and library managers, and added Admins tab.
+- Relabeled "library manager" role to "administrator", and "librarian" role to "user".
 
 ### v0.5.5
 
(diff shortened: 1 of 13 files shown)


tdilauro (approved):
Looks good! 🎸 🙌🏽

PR #17, opened 2021-12-16
Prevent redundant /admin/libraries requests when opening config screen.

ray-lee:
## Description

@jonathangreen observed that when opening the System Configuration screen, two simultaneous requests are sent to /admin/libraries. This ensures that only one request occurs.

This happens because the Header component wants to load the libraries to populate the library selection dropdown, and the Libraries list (which is the first tab, so automatically opens) also wants to load the libraries.

When a libraries fetch starts, the app state is updated, so this just adds checks to both the Header and Libraries components: If a fetch is in flight, another one is not started.

## Motivation and Context

This reduces load on the server, since the libraries request can be expensive to fulfill.

## How Has This Been Tested?

- Log in to the admin UI.
- Open the Network panel in the developer tools.
- Click on System Configuration.

Verify that:
- Only one request is made to /admin/libraries.
- The expected libraries appear in the Library Configuration list.
- The expected libraries appear in the Select a Library dropdown.

Click on other tabs, and verify that the lists loaded properly (because one of the code changes affects all editable lists).

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Changelog
 
-### December 14, 2021
+### December 16, 2021
 
 #### Updated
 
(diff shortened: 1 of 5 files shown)


tdilauro (approved):
Looks good and is definitely doing the job.

tdilauro:
@jonathangreen My local testing showed improvement from > 12s (w/ 2 fetches) to just over 10s when only one active fetch to this endpoint. YMMV, of course.

jonathangreen:
That is great! Thank you for hunting this down @ray-lee!

PR #19, opened 2022-04-21
Remove complaints

ray-lee:
## Description

This removes the Complaints tab from the header and the book details screen.

Before:
complaints-before

After:
complaints-after

## Motivation and Context

We are, at least temporarily, removing the "Report a Problem" feature from the apps so we need to remove the related Complaints feature from the Palace Manager Admin UI.

Notion: https://www.notion.so/lyrasis/Remove-Complaints-feature-from-Palace-Manager-Admin-UI-40d74cce55fc43e0900085cf83cb69c9

## How Has This Been Tested?

Tested by opening the Admin UI, and opening book details screens. The Complaints tab was verified to no longer appear in the header, nor in the book details. This is a view-only change -- actions and reducers that relate to complaints were not removed, to make it easy if we ever want to add this feature back -- so the impact is pretty limited.

Unit tests were updated to no longer expect the Complaints tabs to appear.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 ## Changelog
 
-### December 16, 2021
+### v0.0.8
+
+#### Updated
+
+- Removed Complaints tab from header and book details screen.
+
+### v0.0.7
 
 #### Updated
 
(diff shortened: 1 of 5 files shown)


RishiDiwanTT (approved):
This looks straight forward enough.

PR #20, opened 2022-04-22
Add count to system configuration lists.

ray-lee:
## Description

When editing lists of configuration items in the System Configuration tabs, the count of items in the list is now displayed. The count is shown above the list, on the same line as the Create New button.

Screenshots:

library-count
admin-count
authentication-count
collection-count

## Motivation and Context

The number of configured libraries is commonly requested. Adding the count at the top of the Library Configuration screen in the CM saves time for the Implementation Team.

It seemed useful (or at least not harmful) to have counts on all the lists, so I added the count to the component used for all configuration lists, not just the library list.

## How Has This Been Tested?

I tested by clicking through all the System Configuration tabs, and validated that the count appears, and is correct. A unit test has been added to verify that a count is rendered on lists.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 ## Changelog
 
-### December 16, 2021
+### v0.0.8
+
+#### Updated
+
+- Added display of item count to system configuration lists.
+
+### v0.0.7
 
 #### Updated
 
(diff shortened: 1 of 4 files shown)


RishiDiwanTT (approved):
Looks good

PR #22, opened 2022-04-22
Move Dashboard to be the first link in the header.

ray-lee:
## Description

This moves the Dashboard link in the header to be the first item, instead of being several items deep in the list.

## Motivation and Context

Currently, the Dashboard option is six items deep in the menu bar at the top. It would be much more obvious and convenient for library staff members if Dashboard is moved to be the first option in that list.

Notion: https://www.notion.so/lyrasis/Move-Dashboard-to-the-first-menu-option-available-in-the-tabs-at-the-top-of-screen-1d04bbefc4724b199e4391510b17c4fd

## How Has This Been Tested?

Tested by logging in as a system admin and a librarian, and verifying that Dashboard appears first in the header when a library is selected. When a library is not selected (e.g. on the System Configuration tab), there should be no change.

Unit tests have been updated to expect Dashboard to be the first link.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Removed Complaints tab from header and book details screen.
 - Added display of item count to system configuration lists.
 - Limited visibility of Patrons link in header to system administrators.
+- Moved Dashboard to be the first link in the header.
 
 ### v0.0.7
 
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
LGTM!

PR #23, opened 2022-05-02
Remove circulation links from book details.

ray-lee:
## Description

This removes the circulation links (Borrow, Download, Read buttons) from the book details page. The count of copies available is retained.

I previously mentioned making this change in a fork of [NYPL-Simplified/opds-web-client](https://github.com/NYPL-Simplified/opds-web-client), but I realized that there was a good way to do it in just this repo.

Screenshot:
borrow-removed

## Motivation and Context

This reduces unneeded complexity in the Admin UI, as there is no in-browser reader for EPUB books.

Notion: https://www.notion.so/lyrasis/Hide-Borrow-button-from-Admin-User-accounts-68fbcb4d0a3f44fc942a247a7bc08e40

## How Has This Been Tested?

- Open some book details, using users with different permission levels.
- Verify that for all books and all users, there is no "Borrow", "Download", or "Read" button above the number of copies available.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Changelog
 
+### v0.0.9
+
+#### Updated
+
+- Removed circulation buttons (Borrow, Download, Read) from book details screen.
+
 ### v0.0.8
 
 #### Updated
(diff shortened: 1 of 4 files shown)


RishiDiwanTT (approved):
LGTM!

PR #31, opened 2022-07-21
Refactor CustomListEditor and related components.

ray-lee:
## Description

This refactors the `CustomListEditor`, `CustomListEntriesEditor`, and `CustomListSearch` components to:

- Make them stateless.†
- Remove data manipulation functions that are better implemented outside of the view layer.

The state that was being kept by these components is now lifted up into the Redux store. Several new action creators have been added to represent the actions users can perform on the custom list editor form. A new `customListEditor` reducer handles these actions, and updates the state accordingly. The view components are now pure functions that accept props (pulled from the Redux state), and return a rendered DOM.

† There is still one tiny bit of state in `CustomListEntriesEditor`, the source of a drag while a drag-and-drop is in progress. This can be moved later, if we ever need to generalize drag-and-drop.

This presentation contains an overview of the refactoring: https://docs.google.com/presentation/d/1v-0KdPXHT_fe6suydnUhTeY_NrVFt_lYqFLbPEH9JZc

## Motivation and Context

This greatly simplifies the view components, makes them more testable, and reduces the chance of introducing bugs. It was originally motivated by trying to fix the Save and Cancel buttons being enabled before any changes to a list had been made (https://www.notion.so/lyrasis/Custom-List-Manager-has-active-Save-Cancel-buttons-before-anything-has-been-changed-06ae75f1e064469da0e70bd662396059). This was caused by a confusing interaction between the props and state of `CustomListEditor`. Refactoring these components to remove state allows us to fix this issue, and makes it easier to add new features.

Notion: https://www.notion.so/lyrasis/Refactor-list-editing-components-to-be-stateless-1e9dbc5af08643b99867d4bdabdcb536

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Changelog
 
+### v0.0.11
+
+#### Fixed
+
+- When editing lists, the Save and Cancel buttons are now correctly disabled when a new list is opened, before any changes have been made.
+
 ### v0.0.10
 
 #### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
         "draft-convert": "^2.1.5",
         "draft-js": "0.11.7",
         "font-awesome": "^4.6.3",
+        "immer": "^9.0.15",
         "isomorphic-fetch": "^3.0.0",
         "library-simplified-reusable-components": "1.3.18",
         "numeral": "^2.0.6",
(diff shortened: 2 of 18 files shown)


dbernstein (approved):
@ray-lee:  very impressive work.  It looks good to me, but I must admit I am not familiar with React.

dbernstein (approved):

dbernstein (approved):

PR #37, opened 2022-08-17
Fix styling issues.

ray-lee:
## Description

This fixes a few styling issues:

- When the LastPass browser extension is installed and enabled, a large block of whitespace appears at the top of the screen, making the content invisible without scrolling down. This started happening with a recent update to the LastPass extension, which inserts a top-level `div` in the web page being viewed. The CSS was not expecting this extra `div` to exist.
- The advanced search input fields were misaligned in Firefox.
- There were unnecessary bullets appearing on the list items in the search results and list entries.

## Motivation and Context

The admin UI appears to be completely broken in browsers where the LastPass extension is enabled. The UI is usable after scrolling down to make it visible, but it's hard to discover you have to do that.

Notion: https://www.notion.so/lyrasis/All-CMs-load-with-large-white-space-at-top-of-screen-on-Chrome-and-Edge-8b3e10a667be4783a4a7c18979968904

The other changes just make the UI look better.

## How Has This Been Tested?

- I looked at the UI in Chrome and Firefox, with the LastPass extension enabled, and checked various catalog and configuration screens to make sure the blank space is gone. 
- I checked that the UI still looks good with the LastPass extension disabled.
- I edited a list in Firefox, and verified that the fields in the advanced search section are correctly aligned.
- I did a search in the list editor, and verified that the results list does not have bullets, and that when books are dragged to the entries list, they also do not have bullets.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 ## Changelog
 
+### v0.0.12
+
+#### Fixed
+
+- The screen no longer appears to be blank when the LastPass browser extension is enabled.
+- The advanced search input fields are now correctly aligned in Firefox.
+- Spurious bullets have been removed from the search result and list entry lists.
+
 ### v0.0.11
 
 #### Fixed
--- a/src/stylesheets/advanced_search.scss
+++ b/src/stylesheets/advanced_search.scss
@@ -6,6 +6,15 @@
       gap: 4px;
     }
 
+    .form-group {
+      display: inline-block;
+      margin: 0;
+
+      > div {
+        display: inline-block;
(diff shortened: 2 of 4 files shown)


dbernstein (approved):
Looks good.  I didn't test it manually, but it looks like you did @ray-lee .

tdilauro (approved):
@ray-lee Did you test this withOUT the LastPass extension installed? If so, would you update the PR comment to that effect. I, too, am approving this, since the CSS looks like it will handle both cases properly.

ray-lee:
Yes, I've updated the comment.

PR #38, opened 2022-08-19
Implement UI for automatically updating lists.

ray-lee:
## Description

This implements the UI for saving and loading auto-updating lists. It adds radio buttons to select if a list is auto-updating (now the default), or manually updated. If a list is manually updated, the functionality remains the same as it was prior to this PR. If the list is auto-updating, the list of selected entries to the right of the search results becomes hidden, and the buttons in the search results that are used to manually add entries to the list are removed. The search results become a read-only sample of what the list will contain.

To improve the UX, search results are now automatically updated as search filters are added and removed, and the Search button has been removed. Hopefully this won't overtax the CM, but it can be changed back if it does.

When a list is retrieved from the CM, it may contain a saved search query, which is now loaded into the advanced search builder when the list is opened. A saved search query is automatically executed when a list is opened.

Since the search parameters can now be saved, they now are considered in the calculation of whether or not a list is valid, and whether it has been modified, which they were not previously. They also are now affected by the cancel changes function, which they were not previously. In order to implement this, the search parameters are now stored in the Redux store with a baseline (last saved) state and a current state, similar to other list properties.

This PR also makes a correction to the back-end field name for Distributor search, which was wrong before.

There are a couple of remaining issues to work out:
- Currently only the advanced search filters are saved. The entry point (book, audiobook) and sort parameters are not. There is a ticket open for the CM to support saving these other parameters.
- It looks like a new list can't be initially saved as auto updating (the setting is lost), but it can be after the initial save. I _think_ this is a CM issue, will discuss with Rishi.

Video:

https://user-images.githubusercontent.com/1395885/185525022-7f8f1e47-4e55-4431-a0db-5dfa9aa78e89.mov

## Motivation and Context

This feature has been requested by libraries, to make it easier to maintain lists.

Notion: https://www.notion.so/lyrasis/Admin-UI-Automatically-add-new-titles-to-list-when-they-match-search-parameters-b86ceeffe55d4628b822e661eb992544

## How Has This Been Tested?
- I saved lists with the auto update option selected, and verified that they are loaded correctly from the CM.
- I verified that lists can be changed from auto to manual and vice versa, and correctly saved and loaded either way.
- I verified that search results are automatically updated when filters are changed, and when a new list is opened.

Unit tests are included that cover a lot of the functionality.

## Checklist:


dbernstein (approved):
The code looks good, but,  not being well-versed in React, I don't fully understand how it all works.

tdilauro (approved):
Looks good to me. 🚀

PR #39, opened 2022-08-25
Save search parameters with auto-updating lists.

ray-lee:
## Description

This sends the media type and sort order parameters when saving auto updating lists. Previously, only the advanced search query was sent.

## Motivation and Context

This allows the CM to exactly reproduce the search that was configured by the user when editing a list in the admin UI. In the UI, the user is able to select a media type (all, book, audio) and a sort order (relevance, title, author), in addition to an advanced search query. These all affect the search results. (Somewhat non-intuitively, setting the sort order to title or author changes the search results, because a relevance cut-off is automatically applied.)

This finishes off https://www.notion.so/lyrasis/Admin-UI-Automatically-add-new-titles-to-list-when-they-match-search-parameters-b86ceeffe55d4628b822e661eb992544. (There is some discussion of this in the comments of that ticket).

## How Has This Been Tested?

- Create or edit a list, with auto update set to true.
- Select values for the Search For and Sort By fields in the Search for Titles section.
- Save the list. The selected values should be retained after save.
- Reload the page in the browser. The selected values should be retained after reload.
- Repeat with various combinations of values of Search For and Sort By.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Changelog
 
+### v0.2.1
+
+#### Fixed
+
+- The selected media type and sort order are now correctly saved with auto updating lists.
+
 ### v0.2.0
 
 #### Updated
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
Looks good! 🏆 

Really nice to see these tests where the method being tested takes `state` as an argument.

PR #40, opened 2022-08-29
Add enableAutoList feature flag.

ray-lee:
## Description

This adds an `enableAutoList` feature flag to the application configuration (default false). If the flag is true, the UI for creating automatically-updating lists is enabled. Otherwise, it is hidden. Example usage:

```
var circulationAdmin = new CirculationAdmin({
    csrfToken: "sometoken",
    tos_link_href: "https://thepalaceproject.org/terms-of-service/",
    tos_link_text: "Terms of Service for presenting content through the Palace client applications",
    showCircEventsDownload: true,
    settingUp: false,
    email: "ray.lee@lyrasis.org",
    roles: [{"role": "system" },],
    featureFlags: {
        enableAutoList: true
    }
});
```

This also includes minor simplifications to the state transition functions that are called when lists are loaded. I noticed that I had made them unnecessarily complicated.

## Motivation and Context

This allows the CM to disable the UI for auto-updating lists while back-end support is being completed, and to enable it later without needing to do a new release of the UI. It provides a pattern to do this for other features in the future.

Notion: https://www.notion.so/lyrasis/Hide-auto-list-functionality-until-it-is-working-1a762092f8b947d4a7a4c53a686567a3

## How Has This Been Tested?

I verified the following in the UI:

By default, when editing a custom list:
- The auto update/manual selection radio buttons are not shown at the bottom of the Search for Titles section.
- The list entries column is visible to the right of the search results
- Search results can be dragged from the left column to the right, and the selected entries are saved.
- The above is true for both existing and new lists.
- Lists that were saved on cerberus as auto updating become manually selected when they are opened. (I didn't thoroughly test this scenario where lists that were saved as auto-updating are opened with the feature disabled, since this won't occur in production data).

When the auto update feature is enabled:
- The auto update/manual selection radio buttons are shown at the bottom of the Search for Titles section.
- The list entries column is not shown to the right of search results.
- When new lists are created, they default to auto updating.

Unit tests are included.

## Checklist:


tdilauro (approved):
Looks good! 🎸

PR #41, opened 2022-09-01
Implement global announcements.

ray-lee:
## Description

This adds a Sitewide Announcements configuration tab, where system administrators can configure announcements for all libraries on a CM.

For consistency, this uses the same UI components as the Announcements section in the library configuration screen, and it works identically. I'm not a fan of the UX of those components, but I'm not rewriting them in this PR; that can be a future task. I also noticed some weird but non-blocker bugs in those components, but those can also be addressed in future work.

## Motivation and Context

This allows administrators to send announcements to all users on a CM when needed, for example to announce outages.

Notion: https://www.notion.so/lyrasis/Add-to-front-end-Add-ability-for-super-admins-to-send-mass-announcements-to-all-libraries-configure-36d71b64670f42bc9a8a96d7cca2f025

## How Has This Been Tested?

- Log in as a system administrator.
- On the configuration screen, there should be a Sitewide Announcements tab at the bottom of the list.
- Open the tab.
- Add announcements and save using the Submit button at the bottom. Reload the page in the browser. The saved announcements should be retrieved.
- Edit an announcement (the UX is weird for this, it moves the announcement to the editing area at the bottom). Save and reload, and verify that the edited announcement is retrieved.
- Delete an announcement. Save and reload, and verify that the deleted announcement remains gone.
- Log in as a library administrator.
- The Sitewide Announcements tab should not appear on the configuration screen.
- Log in as a library user.
- The Sitewide Announcements tab should not appear on the configuration screen. 

Unit tests are included.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,14 @@
 ## Changelog
 
-### v0.2.1
+### v0.3.0
 
 #### Fixed
 
 - The selected media type and sort order are now correctly saved with auto updating lists.
+
+#### Updated
+
+- Added Sitewide Announcements configuration tab.
 
 ### v0.2.0
 
--- a/src/__tests__/actions-test.ts
+++ b/src/__tests__/actions-test.ts
@@ -982,4 +982,59 @@ describe("actions", () => {
       expect(data).to.eql(customListDetailsData);
     });
   });
+
+  describe("fetchSitewideAnnouncements", () => {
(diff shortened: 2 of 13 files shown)


tdilauro (approved):
Looks good to me! 🎉

tdilauro:
I ran some local tests of frontend/backend interaction of this and this frontend implementation seems to be working well.

PR #42, opened 2022-09-15
Implement list sharing.

ray-lee:
## Description

This adds support for sharing lists with other libraries on the CM, and for viewing lists shared by other libraries.

- The list sidebar now has filtering options:
  - All: show all lists
  - Owned: show lists owned (created) by this library
  - Shared: show lists owned by this library that are shared with other libraries
  - Subscribed: show lists owned by other libararies, that have been shared with this library
- The list editor now has a Sharing section (only for saved lists) that shows the current sharing status, and contains a button to share a list.
- Lists that are shared now cannot be deleted.
- Lists that are owned by other libraries are now read-only.

This also includes some CSS simplification.

## Motivation and Context

This allows lists to be shared among libraries in a consortium without manually creating the same list over and over.

Notion: https://www.notion.so/lyrasis/Create-interface-for-list-sharing-between-libraries-on-the-same-Palace-Manager-a63f04a22a8043e19b283f0ac5863841

## How Has This Been Tested?

1. Create a new list. There should not be a "Sharing" section at the top of the form.
2. Save the list. The "Sharing" section should appear at the top of the form. In the Sharing section, the status should be "Not  shared", and there should be a "Share" button. In the sidebar, the tile for the list should have "Editing" and "Delete" buttons.
3. Click the Share button. The button should become disabled, and the icon should become a spinner, until the sharing operation is complete. When complete, the status should become "Shared", the button label should become "Share again", and the button should become enabled. In the sidebar, the "Delete" button should disappear, and in the lower right, there should be a sharing icon. A descriptive tooltip should appear when hovering over the sharing icon in the sidebar.
4. At the top of the sidebar, select "Shared" in the Show menu. The shared list should continue to be shown, and any unshared lists should disappear.
5. Select "Subscribed" in the Show menu. All lists owned by this library should disappear.
6. Select "All" in the Show menu. All lists should appear.
7. Change to a different library, and go to the Lists screen.
9. At the top of the sidebar, select "All" in the Show menu. The newly created list should appear, with a sharing icon at the top left (preceding the title). A descriptive tooltip should appear when hovering over the sharing icon. There should be a "View" button, but no Delete button.
10. Open the newly created list. The form should be read-only, and the search results list should not be shown. In the Sharing section, the status should be "Subscribed".
11. At the top of the sidebar, select "Subscribed" in the Show menu. The newly created list should continue to be shown, and any lists owned by the current library should disappear.

## Checklist:

[...]

PR #45, opened 2022-09-21
Fix error loading lists page when there are no owned lists.

ray-lee:
## Description

This fixes a bug in #42. When a library only has lists that it does not own, there is an error opening the Lists screen, and nothing appears.

Currently, when the Lists screen is opened, the first list in the sidebar is automatically opened for editing (or if there are no lists, the form to create a new list is opened). With #42, the lists in the sidebar are by default filtered to show only owned lists, so it's possible for lists to exist, but for none to appear in the sidebar when the Lists screen is first opened. The code that opens the detail view doesn't expect this.

Now, when there are lists, but none are owned, the form to create a new list is opened.

Also, the function that was named `sortedLists` has been changed to `filteredSortedLists` to make it clear that it both filters and sorts lists.

## Motivation and Context

This should have been included in #42, to make it possible to share lists and work with lists shared by other libraries. Notion: https://www.notion.so/lyrasis/Create-interface-for-list-sharing-between-libraries-on-the-same-Palace-Manager-a63f04a22a8043e19b283f0ac5863841

## How Has This Been Tested?

On a CM that includes #42:
1. In a library, create a list and share it.
1. Change to another library that has no lists of its own.
1. Open the Lists screen.
   The Lists sidebar should appear, with "Owned" selected in the "Show" menu. There should be no lists in the sidebar. On the right, the form to create a new list should appear.
1. In the sidebar, select "Subscribed" in the "Show" menu.
   The shared list should appear.

## Checklist:

### DIFF
--- a/src/components/CustomLists.tsx
+++ b/src/components/CustomLists.tsx
@@ -164,7 +164,7 @@ export class CustomLists extends React.Component<
   renderSidebar(): JSX.Element {
     return (
       <CustomListsSidebar
-        lists={this.sortedLists()}
+        lists={this.filteredSortedLists()}
         library={this.props.library}
         identifier={this.props.identifier}
         isLibraryManager={this.context.admin.isLibraryManager(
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Looks good! 🐞

PR #46, opened 2022-09-21
Clear list entries from state when opening a new list.

ray-lee:
## Description

The clears the custom list entries from the Redux store when opening a different list or creating a new list.

Specifically, the `openCustomListEditor` action creator now dispatches the `CUSTOM_LIST_DETAILS_CLEAR` action in addition to the `OPEN_CUSTOM_LIST_EDITOR` action, and a handler for `CUSTOM_LIST_DETAILS_CLEAR` has been implemented that restores the initial (empty) state.

## Motivation and Context

When a list is edited, its entries are loaded, and stored in the Redux store. When a different list is opened, or a new list is created, the entries were not being cleared. This didn't cause an evident problem if a different list was opened, because its entries would be loaded immediately, replacing the old entries. But when a new list is created, there are no entries to load, so the continued existence of the entries from the previous list can cause issues.

This manifested as the "Load more" button on the entries list appearing when creating a new list, even though a new list has no entries. The "Load more" button was left over from the previous list being edited.

Notion: https://www.notion.so/lyrasis/Load-More-button-appears-on-list-entries-when-creating-a-new-list-e07458826c3640539ac8dfd0adb959b8

## How Has This Been Tested?

1. Edit a list that has more than 50 entries. The "Load more" button should appear on the entries list.
2. Click the "Create New List" button at the top of the sidebar.
   On the form for the new list, the "Load more" button should not appear on the entries list.
1. Edit some other lists, and verify that the list entries appear correctly, and the the "Load more" button appears when it is supposed to. 

## Checklist:

### DIFF
--- a/src/__tests__/actions-test.ts
+++ b/src/__tests__/actions-test.ts
@@ -1037,4 +1037,36 @@ describe("actions", () => {
       expect(fetchArgs[0][1].body).to.equal(formData);
     });
   });
+
+  describe("openCustomListEditor", () => {
+    it("dispatches CUSTOM_LIST_DETAILS_CLEAR and OPEN_CUSTOM_LIST_EDITOR", async () => {
+      const dispatch = stub();
+
+      const getState = () => ({
+        editor: {
+          customLists: {
+            data: {
+              foo: "bar",
+            },
+          },
+        },
+      });
+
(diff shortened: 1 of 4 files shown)


tdilauro (approved):
LGTM! 🐛

PR #49, opened 2022-09-28
Fix breadcrumb issues when opening book details from the list editor.

ray-lee:
## Description

This makes a few changes to the breadcrumb trail that is displayed on the book detail page when it is opened from the list editor:
cerberus tpp-qa lyrasistechnology org_admin_web_collection_cerberus-test-1%2Flists%2FPalace%20Marketplace%2Fcrawlable_book_cerberus-test-1%2FISBN%2F9788726607031

- When the book detail page is opened from the list editor while creating a new list, don't show breadcrumbs. Since the new list hasn’t been saved, there is nowhere for the breadcrumb to go.
- When the book detail page is opened from the list editor while editing an existing list, use the saved list name to build the breadcrumb trail, not the current unsaved name. The current name may have been changed, and therefore would not be known to the CM yet.
- While editing an existing list, when the View Details button is clicked from a search result (as opposed to a list entry), don’t show a breadcrumb trail on the book detail page. Showing the list name in the breadcrumb trail implies that the book is in the list, but search results aren’t necessarily in the list.

## Motivation and Context

This fixes an error opening the book detail page from the list editor while creating a new list, before the list has been saved. I found the other related issues while investigating that error.

Notion: https://www.notion.so/lyrasis/Error-appears-when-viewing-a-title-before-a-list-has-been-saved-2b75711414874784af874a8235f59a0c

## How Has This Been Tested?

1. Create a new list. Before saving it, perform a search, and click the View Details button on one of the results.
   The book detail page should open in a new tab, without an error. There should be no breadcrumb trail on the page.
1. Add one or more entries to the list. Click the View Details button on one of the entries.
   The book detail page should open in a new tab, without an error. There should be no breadcrumb trail on the page.
1. Save the list.
2. Click the View Details button on one of the entries.
   The book detail page should open in a new tab, without an error. The name of the list should appear in the breadcrumb trail.
1. Perform a search, and click the View Details button on one of the results.
   The book detail page should open in a new tab, without an error. There should be no breadcrumb trail on the page.
1. Change the name of the list, but don't save it.
2. Click the View Details button on one of the entries.
   The book detail page should open in a new tab, without an error. The saved name of the list (not the unsaved changed name) should appear in the breadcrumb trail.
1. Save the list.
2. Click the View Details button on one of the entries.
   The book detail page should open in a new tab, without an error. The new name of the list should appear in the breadcrumb trail.

## Checklist:

### DIFF
--- a/src/components/__tests__/CustomListEditor-test.tsx
+++ b/src/components/__tests__/CustomListEditor-test.tsx
@@ -247,6 +247,26 @@ describe("CustomListEditor", () => {
     expect(entriesEditor.props().opdsFeedUrl).to.equal(undefined);
   });
 
+  it("sets the feed url in the entries editor to undefined when the saved list name is falsy", () => {
+    wrapper.setProps({
+      savedName: "",
+    });
+
+    let entriesEditor;
+
+    entriesEditor = wrapper.find(CustomListEntriesEditor);
+
+    expect(entriesEditor.props().opdsFeedUrl).to.equal(undefined);
+
+    wrapper.setProps({
+      savedName: null,
+    });
+
+    entriesEditor = wrapper.find(CustomListEntriesEditor);
+
+    expect(entriesEditor.props().opdsFeedUrl).to.equal(undefined);
+  });
+
   it("shows collections", () => {
     const collectionsPanel = wrapper.find("#add-from-collections-panel");
     const inputs = collectionsPanel.find(EditableInput);
--- a/src/components/__tests__/CustomListEditor-test.tsx
+++ b/src/components/__tests__/CustomListEditor-test.tsx
@@ -224,6 +224,29 @@ describe("CustomListEditor", () => {
     );
   });
 
+  it("sets the feed url in the entries editor to the url for the saved list name", () => {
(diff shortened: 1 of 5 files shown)


src/components/__tests__/CustomListEditor-test.tsx
```diff
@@ -224,6 +224,29 @@ describe("CustomListEditor", () => {
...
+    wrapper.setProps({
+      listId: null, // new, unsaved lists have null listId
+      savedName: "Books to Read Today",
+    });
+
+    const entriesEditor = wrapper.find(CustomListEntriesEditor);
+
+    expect(entriesEditor.props().opdsFeedUrl).to.equal(undefined);
+  });
+
```
tdilauro, line 249: Do you want a test here for when `savedName` is empty string or null?

tdilauro (approved):
May need a test, but otherwise looks good.

src/components/__tests__/CustomListEditor-test.tsx
ray-lee: Yes, added.

PR #50, opened 2022-09-29
Add list filtering to lane editor.

ray-lee:
## Description

This adds the ability to filter the lists that are shown when editing a lane. The options are to show only the lists owned by the current library (the default), only lists shared by other libraries, or all lists.

localhost_8080_admin_web_lanes_ray_create_179

It also adds a share icon to lists that are owned by other libraries, to make them easy to differentiate.

localhost_8080_admin_web_lanes_ray

## Motivation and Context

This makes it easier for librarians to find their own lists, vs. lists shared by other libraries, when creating lanes. 

Notion: https://www.notion.so/lyrasis/Add-filtering-options-to-lists-when-editing-a-lane-08b679fb44124a6aa9cd8743a003a43c

## How Has This Been Tested?

1. In one library, create a list, and share it.
2. In a second library, create a list that is not shared.
3. In the second library, create a new Lane.
   In Available Lists, the Show menu should have "Owned" selected. Only the owned list should appear. The list shared by the first library should not appear.
1. In the Show menu, select "Subscribed".
   The owned list should disappear, and the list shared by the first library should appear.
1. In the Show menu, select "All".
   All the lists should appear. The list shared from the first library should have a share icon next to its name, and the owned list should not.
1. Drag the shared list to the lane on the right.
   The list should disappear from the Available Lists list, and appear in the Lists in This Lane list.
1. In the Show menu, select "Subscribed".
   The Available Lists list should remain empty, and the shared list should continue to appear in the Lists in This Lane list.
1. Drag the shared list from the right to the left.
   The shared list should appear in the Available Lists list, and disappear from the Lists in This Lane list.

## Checklist:

- [x] I have updated the documentation accordingly. https://www.notion.so/lyrasis/Admin-UI-Lane-Creation-a961c06a77f545b2a8814c0012d24ea8

### DIFF
--- a/src/components/LaneCustomListsEditor.tsx
+++ b/src/components/LaneCustomListsEditor.tsx
@@ -2,6 +2,7 @@ import * as React from "react";
 import { DragDropContext, Droppable, Draggable } from "react-beautiful-dnd";
 import { CustomListData } from "../interfaces";
 import AddIcon from "./icons/AddIcon";
+import ShareIcon from "./icons/ShareIcon";
 import TrashIcon from "./icons/TrashIcon";
 import GrabIcon from "./icons/GrabIcon";
 import EditableInput from "./EditableInput";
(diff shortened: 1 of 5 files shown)


src/components/LaneCustomListsEditor.tsx
```diff
@@ -75,7 +118,12 @@ export default class LaneCustomListsEditor extends React.Component<
                             >
                               <GrabIcon />
```
tdilauro, line 119: ll. 119-131 is duplicated below in ll. 188-200. Would it be useful to refactor?

tdilauro (approved):
Looks good! 🚀

src/components/LaneCustomListsEditor.tsx
ray-lee: Good point. I extracted the duplicate code into its own function.

PR #55, opened 2022-11-16
Use Palace fork of the OPDS client library.

ray-lee:
## Description

This replaces the `opds-web-client` dependency with `@thepalaceproject/web-opds-client`, which is our fork built from https://github.com/ThePalaceProject/web-opds-client.

At this time, the two libraries are identical, so there is no functional change.

## Motivation and Context

This prepares us to be able to modify the code of the underlying OPDS client that is used by this project. I'm doing this in preparation for removing the Borrow button from the catalog list view (Notion: https://www.notion.so/lyrasis/Hide-Borrow-button-when-viewing-titles-in-list-view-after-opening-a-lane-in-the-Collection-Manager-b00a7e7bac244b57a6abf2a737f42510), but it will surely be useful for other features and bug fixes in the future.

## How Has This Been Tested?

All unit tests should pass, and the app should continue to work identically. I spot checked various functions in the UI.

## Checklist:


src/stylesheets/global.scss
```diff
@@ -87,7 +87,7 @@ button.btn.inverted, a.btn.inverted {
   }
 }
 
-// Overwrite the rule from opds-web-client that sets all top-level divs to 100% height. This works
+// Overwrite the rule from web-opds-client that sets all top-level divs to 100% height. This works
```
RishiDiwanTT, line 90: I'm curious about why the name has changed around the words.

RishiDiwanTT (approved):
This all seems good.

Except for one problem which I doubt is part of this change, using the `--backend-env=http://localhost:6500`, I get the following error
image

If this issue isn't part of this PR we can address it elsewhere

src/stylesheets/global.scss
tdilauro: @RishiDiwanTT This is because we renamed the project when we forked it. You can see it in the change on l. 62 of the README.

RishiDiwanTT: Oh yes I was quite sure the rename was intentional I was just curious as to the motivation. This is a purely educational question (for myself) and has no bearing on the already approved PR.

ray-lee:
@RishiDiwanTT Yes, that search error is a pre-existing problem that is unrelated. I've been meaning to fix it.

PR #56, opened 2022-11-18
Remove borrow button from catalog list view.

ray-lee:
## Description

This removes the Borrow button from the catalog list view. It upgrades opds-web-catalog to 1.0.0, which allows us to show or hide the circulation links by setting a prop.

Before:
Screen Shot 2022-11-18 at 10 45 58 AM

After:
Screen Shot 2022-11-18 at 10 46 28 AM

## Motivation and Context

The circulation functions are complicated to support in the admin UI, and there is no built-in reader, so we are removing them.

Notion: https://www.notion.so/lyrasis/Hide-Borrow-button-when-viewing-titles-in-list-view-after-opening-a-lane-in-the-Collection-Manager-b00a7e7bac244b57a6abf2a737f42510

## How Has This Been Tested?

In the catalog, click on a lane, and click the list view button at the top. No Borrow buttons should appear in the list, as in the above screenshot. 

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Changelog
 
+### v0.4.2
+
+#### Updated
+
+- The circulation buttons (e.g. Borrow, Read) have been removed from the catalog list view.
+
 ### v0.4.1
 
 #### Fixed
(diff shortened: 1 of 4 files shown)


RishiDiwanTT (approved):
Succinct :+1:

PR #66, opened 2023-01-20
Set up Jest/React Testing Library tests.

ray-lee:
## Description

This adds configuration needed to start writing tests using Jest and React Testing Library (instead of Mocha and Enzyme). Existing Mocha/Enzyme tests are unchanged. New Jest/RTL tests should be placed in the tests/jest directory. The `npm test` script now runs both the old Mocha/Enzyme tests as well as the new Jest/RTL tests.

## Motivation and Context

The Enzyme test utility for React [doesn't currently support React 17](https://github.com/enzymejs/enzyme/issues/2429), and [likely won't support React 18](https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl). In order to advance beyond React 16, we'll need an alternative for testing. The React team is [recommending React Testing Library](https://reactjs.org/docs/testing.html#tools) as a replacement. A common pattern for writing tests using React Testing Library is to [mock components that are not under test](https://reactjs.org/docs/testing-recipes.html#mocking-modules) using Jest's mock function, so Jest is included. While it's not necessary to use Jest as a test runner, mixing Mocha (as a test runner) and Jest (only for the mock function) would be confusing, so the new tests are configured to run using Jest.

## How Has This Been Tested?

- `npm install` in a clean directory (no lib, no node_modules) should work.
- `npm test` should work.

## Checklist:


tdilauro (approved):
Looks good! 🚀

PR #69, opened 2023-01-27
Clear role checkboxes after saving a new user.

ray-lee:
## Description

This fixes a bug where role checkboxes remained checked after saving a new user in the Admins tab. The code stores references to the checkbox components on the form, and clears them after save, but this only worked if there was a single library. If there is more than one library, the code only kept references to the checkboxes for the last library in the list, so only those checkboxes were cleared. The fix stores references to the checkboxes for all libraries, so they can all be cleared.

(The whole notion of storing references to the checkboxes and imperatively calling clear() on them is pretty cringey, but I'm not doing a big refactor now.)

In order to simplify writing a test for the fix, a new test utility function has been added that automatically wraps some JSX in a `ContextProvider` component, so that the component under test can be rendered (and re-rendered) with the appropriate context.

Finally, some typing tweaks were made to the `ContextProvider` component to resolve typing and linter errors.

## Motivation and Context

When creating multiple users, the previously created user's roles were retained for the next user, which made it easy to accidentally assign incorrect roles.

Notion: https://www.notion.so/lyrasis/Admin-Roles-check-boxes-do-not-automatically-reset-after-page-reload-on-CM-4a12a18e7a914eaebfb25756234a3d7a

## How Has This Been Tested?

- On the System Configuration/Admins tab, click the "Create new individual admin" button.
- Enter an email, and select some roles using the checkboxes under "Admin Roles".
- Click the Submit button.
   Expect: All of the role checkboxes are unchecked following the save.
- Repeat for various combinations of roles.

## Checklist:

### DIFF
--- a/src/components/IndividualAdminEditForm.tsx
+++ b/src/components/IndividualAdminEditForm.tsx
@@ -44,8 +44,8 @@ export default class IndividualAdminEditForm extends React.Component<
   private systemRef = React.createRef<EditableInput>();
   private managerAllRef = React.createRef<EditableInput>();
   private librarianAllRef = React.createRef<EditableInput>();
-  private libraryManagerRef = React.createRef<EditableInput>();
-  private librarianRef = React.createRef<EditableInput>();
+  private libraryManagerRefs = {};
+  private librarianRefs = {};
 
   constructor(props) {
     super(props);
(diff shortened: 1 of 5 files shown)


src/components/IndividualAdminEditForm.tsx
```diff
@@ -197,7 +198,11 @@ export default class IndividualAdminEditForm extends React.Component<
                       type="checkbox"
                       disabled={this.isDisabled("manager", library.short_name)}
                       name={"manager-" + library.short_name}
-                      ref={this.libraryManagerRef}
+                      ref={(componentInstance) => {
+                        this.libraryManagerRefs[
+                          library.short_name
+                        ] = componentInstance;
+                      }}
```
ray-lee, line 205: prettier sometimes does some questionable formatting.

tdilauro (approved):
Looks good! 🎸🤘🏽

PR #72, opened 2023-02-07
Remove CDN Config

jonathangreen:
## Description

This completes the work in https://github.com/ThePalaceProject/circulation/pull/785 to remove CDN settings from the admin interface. We are going to take a different approach to using a CDN, and don't require storing the CDN information in the database.

## Motivation and Context

CDN configuration endpoints have been removed from the circulation code, so the admin interface needs to be updated to remove the UI for CDN configuration.

## How Has This Been Tested?

- Tested locally with `npm run dev-server -- --env=backend=https://minotaur.dev.palaceproject.io`
- Ran unit tests

## Checklist:


ray-lee (approved):
Looks good to me.

PR #73, opened 2023-02-09
Mark password field as required

io7m:
## Description

The CM now considers passwords to be required and not optional. This updates the UI so that the password field shows the "Required" text.

## Motivation and Context

See: https://github.com/ThePalaceProject/circulation/pull/801

## How Has This Been Tested?

I just checked the form in the web interface. All of the real changes happened in the CM.

## Checklist:


io7m:
Eyeing test failure.

tdilauro (approved):
This looks good! 🎸 

Ideally, we'd be disabling the the `Submit` button when the updates are not valid, but I think that's more involved and should not be part of this work.

PR #74, opened 2023-02-10
Allow automatically generated lanes to be renamed.

ray-lee:
## Description

This allows automatically generated lanes to be renamed.

- In the Lanes sidebar, every lane now has an edit button, whether it's custom or automatically generated.
- In the lane editor, automatically generated lanes can only be renamed and made visible/invisible. Other functions are not available: deleting, adding/removing lists, and inheriting restrictons from the parent.

The included tests contain examples of using Jest mocks with React Testing Library to achieve something similar to Enzyme's shallow rendering, which allows isolating tests to a single component.

## Motivation and Context

Libraries have requested the ability to rename the automatically generated lanes. Notion: https://www.notion.so/lyrasis/Admin-UI-Add-ability-to-rename-a-default-lane-name-in-the-CM-4b42fd3a308947a7907442d176447de2

## How Has This Been Tested?

- Run using cerberus as a back end: `npm run dev-server -- --env=backend=https://cerberus.staging.palaceproject.io`
- Select Cerberus Test Library 1, and click Lanes
- In the sidebar, click the Edit button on an automatically generated lane (e.g. Fiction, Nonfiction, Young Adult Fiction, etc.)
The lane editor should open. There should be an Edit Name button, a Hide/Show lane button, and a Save button. There should not be a Delete button. There should not be an Available Lists list, nor a Lists in This Lane list. If this is a sublane, there should not be an "Inherit restrictions from parent lane" checkbox.
- Change the name of the list and save.
The new list name should appear in the sidebar, and in the lane editor.
- Navigate to the catalog, and browse to the lane. It should appear with the new name. (It sometimes takes a while for the change to appear on the catalog, which I assume is some kind of caching in the CM).

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - A placeholder and help text are now displayed when the Publication Date field is selected in advanced search.
 - Remove CDN configuration screens.
+- Automatically generated lanes can now be renamed.
 
 #### Fixed
 
--- a/src/components/Lane.tsx
+++ b/src/components/Lane.tsx
@@ -47,8 +47,7 @@ export default class Lane extends React.Component<LaneProps, LaneState> {
   render(): JSX.Element {
     const { lane, active, snapshot, provided, renderLanes } = this.props;
     const hasSublanes = lane.sublanes && !!lane.sublanes.length;
-    const hasCustomLists =
-      lane.custom_list_ids && !!lane.custom_list_ids.length;
+
     return (
       <li key={lane.id}>
         <div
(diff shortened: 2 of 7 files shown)


RishiDiwanTT (approved):
This looks good and functions as described.
Two little feedback points
- There is a flicker while the lanes load after a save under the loading bar, where the edit columns are seen and then disappear once the load is done, maybe default to have hidden edit columns that appear after the load?
- The user has no information as to why they cannot edit these lanes, maybe have a little info panel detailing the special treatment for these lanes inside the RHS panel?

These are subjective feedback points, so I am approving it anyways, they can be implemented or left as is, they do not affect functionality.

ray-lee:
@RishiDiwanTT, Those are good suggestions. I've added an explanation to the lane editor to say that the contents of automatically generated lanes can't be edited. I've also made it so that nothing is displayed in the list editor area while a lane is being loaded for editing, which should resolve the flicker.

RishiDiwanTT (approved):
Looks good :+1:

PR #81, opened 2023-03-26
Adds a RootState interface to reflect complete Redux store state.

tdilauro:
## Description

- Adds a RootState interface to reflect actual Redux store state.
- Introduces a Redux `Provider` container and ensures that both it and the existing `ContextProvider` container, which it now wraps, provide the same store.

## Motivation and Context

The `State` interface that was previously propagated reflected only the app-specific (so-called "editor") state, but did not include the state pulled in from the [web-opds-client](https://github.com/ThePalaceProject/web-opds-client) package (the "catalog" state). So that interface could not fully type the contents of the store. This mismatch also led to developer confusion.

The Redux `Provider` container has been added to enable us to start using the `useDispatch` and `useSelector` Redux hooks and custom hooks based on them. This will allow us to incrementally away from using the existing `ContextProvider` container to propagate the Redux store, which is an unconventional pattern in its own right.

## How Has This Been Tested?

- Spot checked functionality in local development environment.
- Tests pass locally.
[CI tests for this branch](https://github.com/ThePalaceProject/circulation-admin/actions/runs/4522522907/jobs/7965053335) passed.

## Checklist:

### DIFF
--- a/src/components/AccountPage.tsx
+++ b/src/components/AccountPage.tsx
@@ -2,18 +2,18 @@ import * as React from "react";
 import { Store } from "redux";
 import * as PropTypes from "prop-types";
 import ChangePasswordForm from "./ChangePasswordForm";
-import { State } from "../reducers/index";
+import { RootState } from "../store";
 import Header from "./Header";
 import Footer from "./Footer";
 import title from "../utils/title";
 
 export interface AccountPageContext {
-  editorStore: Store<State>;
+  editorStore: Store<RootState>;
   csrfToken: string;
 }
 
 /** Page for configuring account settings. */
-export default class AccountPage extends React.Component<{}, {}> {
+export default class AccountPage extends React.Component<object, object> {
   context: AccountPageContext;
 
   static contextTypes: React.ValidationMap<AccountPageContext> = {
--- a/src/components/ContextProvider.tsx
+++ b/src/components/ContextProvider.tsx
@@ -1,9 +1,8 @@
 import * as React from "react";
 import { Store } from "redux";
 import * as PropTypes from "prop-types";
-import buildStore from "../store";
+import buildStore, { RootState } from "../store";
 import { FeatureFlags, PathFor } from "../interfaces";
-import { State } from "../reducers/index";
 import Admin from "../models/Admin";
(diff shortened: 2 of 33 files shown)


src/components/ContextProvider.tsx
```diff
@@ -25,13 +25,13 @@ export interface ContextProviderProps extends React.Props<ContextProvider> {
...
 > {
-  store: Store<State>;
+  store: Store<RootState>;
   admin: Admin;
   pathFor: PathFor;
 
   constructor(props) {
     super(props);
-    this.store = buildStore();
+    this.store = props.store ?? buildStore();
```
tdilauro, line 34: Key change: Allow our `ContextProvider` to accept an injected store, so that it can have the same one as the Redux `Provider` component. This makes is possible to incrementally transition to the store from the `Provider` without breaking things.

tdilauro (commented):
The significant changes in this PR are discussed in these comments.

The remaining changes are just updating from the previous `State` interface, which described only part of the store state, with the new `RootState` interface.

src/components/AccountPage.tsx
```diff
@@ -2,18 +2,18 @@ import * as React from "react";
...
 import ChangePasswordForm from "./ChangePasswordForm";
-import { State } from "../reducers/index";
+import { RootState } from "../store";
 import Header from "./Header";
 import Footer from "./Footer";
 import title from "../utils/title";
 
 export interface AccountPageContext {
-  editorStore: Store<State>;
+  editorStore: Store<RootState>;
```
ray-lee, line 11: Not a blocker, but `ContextProvider` is naming this `editorStore`, which may lead one to believe that it is just the "editor" half of the store... which I guess is what someone thought it was when they typed it the way they did. So maybe `rootStore` would be more appropriate. But that would require a bunch of changes, and this is still an improvement.

ray-lee (approved):
This looks good!

src/components/AccountPage.tsx
tdilauro: Thanks, @ray-lee. I saw that, which I should've mentioned. The risk mitigator here, based on my sampling, is that the components that are using that name really are using only the `editor` portion of the store. I am envisioning that most, if not all of these should eventually be eliminated by introducing slices, which I believe is a much nicer and more maintainable approach.

PR #82, opened 2023-04-12
Dashboard statistics improvements [requires circulation #993]

tdilauro:
## Description

- Breaks out counts for unlimited license and metered license books.
- Shows count of books available for borrowing.
- Shows counts for owned and available metered licenses.
- Shows popup with explanatory text and exact count for counts in first three groups (patron, circulation, and inventory counts for the single library or as a summary for all authorized libraries).
- Shows popup for collection-level inventory for any collection in the fourth (collection) group.

There are also a couple of non-functional changes of note:
- `Stats` component changed from Redux connected component to functional component with a custom hook, removing deprecated lifecycle methods in the process.
- Moves Redux store dependency from `ContextProvider` component to Redux's `Provider` component.

**WARNING**: Requires the changes in [circulation #993](https://github.com/ThePalaceProject/circulation/pull/993).

## Motivation and Context

Unlimited license (i.e., not Open Access (OA) and not metered license) book counts were not reflected in the metrics displayed in the statistics dashboard. [[Notion](https://www.notion.so/lyrasis/OPDS-2-Imported-Collections-are-not-counted-in-Dashboard-ba4f0907268a4547933f3cd3a849585d?pvs=4)]

## How Has This Been Tested?

- Existing test were re-worked to accommodate new approach.
- Manually ran with circulation PR #993 environment in local development environment.
- [CI tests for feature branch](https://github.com/ThePalaceProject/circulation-admin/actions/runs/4678634144) pass.
- NB: I probably need some `jest/RTL`-based tests, in addition to the existing re-worked tests; but, I wanted to get this into a PR sooner rather than later. Comments to this effect are welcome.

## Checklist:


src/components/DashboardPage.tsx
```diff
@@ -43,7 +43,7 @@ export default class DashboardPage extends React.Component<DashboardPageProps> {
       <div className="dashboard">
         <Header />
         <main className="body">
-          <Stats store={this.context.editorStore} library={library} />
+          <Stats library={library} />
```
tdilauro, line 46: `Stats` now gets specific state from the store in its own custom hook.

src/components/Stats.tsx
```diff
@@ -1,118 +1,100 @@
...
-    fetchLibraries: () => dispatch(actions.fetchLibraries()),
-  };
-}
+  return { ...statistics, libraries, summaryStatistics };
+};
+
+const useStatistics = (): StatsLocalState => {
+  const data = useSelector(statsStateSelector.data);
+  const isLoaded = useSelector(statsStateSelector.isLoaded);
+  const fetchError = useSelector(statsStateSelector.fetchError);
```
tdilauro, line 89: This custom hook gets the needed bits of state from the Redux store...

src/components/Stats.tsx
```diff
@@ -1,118 +1,100 @@
...
+
+  const dispatch = useDispatch();
+  useEffect(() => {
+    const actions = new ActionCreator();
+    dispatch(actions.fetchStatistics());
+  }, []);
```
tdilauro, line 95: ... and picks up the fetch previously performed in the `componentWillMount` lifecycle method.

src/components/LibraryStats.tsx
```diff
@@ -1,180 +1,294 @@
...
+    const newPayload = [
+      ...aboveTheLine,
+      {}, // blank line
+      { value: ">>> Additional Information <<<", color: belowTheLineColor },
+      ...belowTheLine,
+    ];
+
+    // We render the default, but with our overridden payload.
+    return <DefaultTooltipContent {...props} payload={newPayload} />;
+  };
```
tdilauro, line 107: This bit overrides the tooltip to display inventory details for each of the individual collections in the bar chart. It's kind of messy, so would especially appreciate any suggestions for cleaning it up.

src/components/LibraryStats.tsx
```diff
@@ -1,180 +1,294 @@
...
+      ) : (
+        <h2>Statistics for All Libraries</h2>
+      )}
+      <ul className="stats">
+        <li className="stat-group">{patronsGroup(patrons)}</li>
+        <li className="stat-group">{circulationsGroup(patrons)}</li>
+        <li className="stat-group">{inventoryGroup(inventory)}</li>
+        <li className="stat-group stat-group-wide">
+          {collectionsGroup(chartItems, CustomTooltip)}
+        </li>
```
tdilauro, line 122: I factored these out into functions because nesting them inline made it harder to understand what is happening. I wasn't sure whether it would be better to turn them into components, but it seemed easier to call the functions and not have to deal with the props. They aren't used anywhere else.

src/store.ts
```diff
@@ -6,11 +6,16 @@ import {
...
+import thunk from "redux-thunk";
+import catalogReducers from "@thepalaceproject/web-opds-client/lib/reducers/index";
 import { State as CatalogState } from "@thepalaceproject/web-opds-client/lib/state";
 import editorReducers, { State as EditorState } from "./reducers/index";
 
-import thunk from "redux-thunk";
+export interface CombinedState {
+  editor: EditorState;
+  catalog: CatalogState;
+}
```
ray-lee, line 18: Any reason to have both `CombinedState` and `RootState`?

tdilauro: No. No good reason, anyway. 😅

src/components/Stats.tsx
```diff
@@ -1,118 +1,100 @@
...
-    }
-    if (this.props.fetchLibraries()) {
-      this.props.fetchLibraries();
-    }
-  }
-}
+  const collectionIdMap = Object.assign(
+    {},
+    ...statistics.collections.map((c) => ({ [c.id]: c }))
+  );
```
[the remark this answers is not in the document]
tdilauro: Happy to use the approach that is most intuitive.

src/components/LibraryStats.tsx
tdilauro: I'm overriding only the `payload` here, so I'll need the rest of the `props` to pass on to the default tooltip. So, I'll keep this as a React component and reference it properly on the chart `<Tooltip />`.

tdilauro: I was previously referencing something inside the closure, which is why it was there; but that is no longer the case.

PR #84, opened 2023-06-28
Add language=all param to custom list searches.

ray-lee:
## Description

This adds the `language=all` query parameter to all custom list searches.

## Motivation and Context

Only English titles were being returned in custom list search results.

JIRA: https://ebce-lyrasis.atlassian.net/browse/PP-80

## How Has This Been Tested?

- Run `npm run dev-server -- --env=backend=https://minotaur.dev.palaceproject.io`
- Go to http://localhost:8080/admin
- Log in using minotaur credentials
- Open the Lists screen for Minotaur Test Library
- Create a new list
- In Works to Include, search for _title contains jane eyre_
A French title ("Jane Eyre ou Les Mémoires d'une institutrice") and a German title ("Jane Eyre, die Waise von Lowood") should be included in the results.
- Add _language equals french_ to the search filters
The results should now contain only the French title.
- Delete the _language equals french_ filter
The English and German titles should reappear in the results.

## Checklist:

### DIFF
--- a/src/reducers/__tests__/customListEditor-test.ts
+++ b/src/reducers/__tests__/customListEditor-test.ts
@@ -2714,7 +2714,7 @@ describe("custom list editor reducer", () => {
             entryPoint: "Book",
             terms: "foo bar baz",
             sort: "title",
-            language: "eng",
+            language: "all",
             advanced: {
               include: {
                 query: null,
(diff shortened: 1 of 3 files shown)


tdilauro (commented):
This looks good. Is there a test that covers the case of overriding the new default language?

ray-lee:
I need to rework this. Moving back to draft.

ray-lee:
Ready for review. I added a test to show what should happen when a user adds a language filter: The user's filter is added to the JSON query, but the `language` parameter is still set to `all` to prevent the query results from being filtered.

tdilauro (approved):
Looks good to me! 🥇

PR #93, opened 2023-10-13
Setup admin ui to publish a next tag on commit to main (PP-164)

jonathangreen:
## Description

This will tag a pre-release version of the cm admin interface on every push to main under the next tag.

- Create a new `test-and-tag.yml` workflow run on pushes to main.
- Modify `test.yml` to only run on pull requests.

## Motivation and Context

This will allow us to setup the CM to use the next tag in our development server, so that we get up to date versions of the CM admin interface in that environment.

## How Has This Been Tested?

- I tested this as far as I could in my own fork of the repo, but I wasn't able to test the actual npm release process, so this may need a bit of tweaking once its merged.

## Checklist:


tdilauro (approved):
👍🏽

PR #95, opened 2023-11-02
PP-668 Removed the analytics service

RishiDiwanTT:
## Description
Removed the analytics service tabs from the System Configuration page.
These are no longer required as the CM will now configure the analytics without user interaction.

:warning: Unit tests did not run successfully locally, but neither does the main branch.

## Motivation and Context
The refactored analytics on the CM no longer require the Admin UI to control the "settings".

## How Has This Been Tested?
Manually tested the removal and lack of API requests to the CM API.
This only affects the System Configuration page.

## Checklist:


jonathangreen (approved):
Looks good to me!

PR #97, opened 2023-12-02
Add quicksight dashboard

dbernstein:
## Description

```
nvm install 18
nvm use 18
npm run dev-server -- --env=backend=https://minotaur.dev.palaceproject.io
```
Once you've logged in, to see the dashboard you need to go to:

http://localhost:8080/admin/web/quicksight

The URL is not exposed in the App (since I don't think it is quite ready for prime time).

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-517

## How Has This Been Tested?

Added unit test.
Tested it manually.
## Checklist:

### DIFF
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -26,7 +26,7 @@ import {
   DiagnosticsData,
   FeatureFlags,
   SitewideAnnouncementsData,
-  StatisticsData,
+  StatisticsData, DashboardURIData,
 } from "./interfaces";
 import { CollectionData } from "@thepalaceproject/web-opds-client/lib/interfaces";
 import DataFetcher from "@thepalaceproject/web-opds-client/lib/DataFetcher";
@@ -190,6 +190,8 @@ export default class ActionCreator extends BaseActionCreator {
   static readonly RESET_ADOBE_ID = "RESET_ADOBE_ID";
 
   static readonly DIAGNOSTICS = "DIAGNOSTICS";
+
+  static readonly DASHBOARD_URI : "DASHBOARD_URI"
 
   csrfToken: string;
 
@@ -1066,4 +1068,14 @@ export default class ActionCreator extends BaseActionCreator {
       value,
     };
   }
+
+  fetchDashboardUri(dashboardId: string) {
+    const url = "/admin/quicksight_embed/" + dashboardId;
+    return this.fetchJSON<DashboardURIData>(
+        ActionCreator.DASHBOARD_URI,
+        url
+    ).bind(this);
+  }
+
 }
+
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -26,7 +26,8 @@ import {
   DiagnosticsData,
   FeatureFlags,
   SitewideAnnouncementsData,
-  StatisticsData, DashboardURIData,
+  StatisticsData,
+  DashboardURIData,
 } from "./interfaces";
 import { CollectionData } from "@thepalaceproject/web-opds-client/lib/interfaces";
 import DataFetcher from "@thepalaceproject/web-opds-client/lib/DataFetcher";
@@ -190,8 +191,7 @@ export default class ActionCreator extends BaseActionCreator {
   static readonly RESET_ADOBE_ID = "RESET_ADOBE_ID";
 
   static readonly DIAGNOSTICS = "DIAGNOSTICS";
-
-  static readonly DASHBOARD_URI : "DASHBOARD_URI"
+  static readonly DASHBOARD_URI: "DASHBOARD_URI";
 
   csrfToken: string;
 
@@ -1072,10 +1072,8 @@ export default class ActionCreator extends BaseActionCreator {
   fetchDashboardUri(dashboardId: string) {
     const url = "/admin/quicksight_embed/" + dashboardId;
     return this.fetchJSON<DashboardURIData>(
-        ActionCreator.DASHBOARD_URI,
-        url
+      ActionCreator.DASHBOARD_URI,
+      url
     ).bind(this);
   }
-
 }
-
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -190,7 +190,7 @@ export default class ActionCreator extends BaseActionCreator {
   static readonly RESET_ADOBE_ID = "RESET_ADOBE_ID";
 
   static readonly DIAGNOSTICS = "DIAGNOSTICS";
-  static readonly QUICKSIGHT_EMBEDDED_URL: "QUICKSIGHT_EMBEDDED_URL";
+  static readonly QUICKSIGHT_EMBEDDED_URL = "QUICKSIGHT_EMBEDDED_URL";
 
   csrfToken: string;
 
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -26,7 +26,7 @@ import {
   FeatureFlags,
   SitewideAnnouncementsData,
   StatisticsData,
-  DashboardURIData,
+  QuickSightEmbeddedURLData,
 } from "./interfaces";
 import { CollectionData } from "@thepalaceproject/web-opds-client/lib/interfaces";
 import DataFetcher from "@thepalaceproject/web-opds-client/lib/DataFetcher";
@@ -190,7 +190,7 @@ export default class ActionCreator extends BaseActionCreator {
   static readonly RESET_ADOBE_ID = "RESET_ADOBE_ID";
 
   static readonly DIAGNOSTICS = "DIAGNOSTICS";
-  static readonly DASHBOARD_URI: "DASHBOARD_URI";
+  static readonly QUICKSIGHT_EMBEDDED_URL: "QUICKSIGHT_EMBEDDED_URL";
 
   csrfToken: string;
 
@@ -1068,11 +1068,11 @@ export default class ActionCreator extends BaseActionCreator {
     };
   }
 
-  fetchQuicksightEmbedUri(dashboardId: string, ld: LibrariesData) {
+  fetchQuicksightEmbedUrl(dashboardId: string, ld: LibrariesData) {
     const library_uuids: string = ld.libraries.map((l) => l.uuid).join(",");
     const url = `/admin/quicksight_embed/${dashboardId}?libraryUuids=${library_uuids}`;
-    return this.fetchJSON<DashboardURIData>(
-      ActionCreator.DASHBOARD_URI,
+    return this.fetchJSON<QuickSightEmbeddedURLData>(
+      ActionCreator.QUICKSIGHT_EMBEDDED_URL,
       url
     ).bind(this);
   }
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -22,7 +22,6 @@ import {
   RightsStatusData,
   CatalogServicesData,
   SelfTestsData,
-  PatronData,
   DiagnosticsData,
   FeatureFlags,
   SitewideAnnouncementsData,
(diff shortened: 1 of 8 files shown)


tdilauro (commented):
I wasn't entirely sure if this PR was ready for review yet, but I have a few comments/suggestions that I hope can be helpful in any case.

src/components/QuicksightDashboardPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+
+  render(): JSX.Element {
+    const { library } = this.props.params;
+    return (
+      <div className="quicksight-dashboard">
+        <Header />
+        <main className="body">
+          <QuicksightDashboard dashboardId="library"/>
```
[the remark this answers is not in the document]
dbernstein: It's not. The dashboard is the library / librarian dashboard view.

src/index.tsx
```diff
@@ -87,6 +88,10 @@ class CirculationAdmin {
                 path="/admin/web/dashboard(/:library)"
                 component={DashboardPage}
               />
+              <Route
+                  path="/admin/web/dashboard/quicksight(/:library)"
+                  component={QuicksightDashboardPage}
+              />
```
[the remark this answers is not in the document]
dbernstein: Okay - I'll fix that.

src/components/QuicksightDashboard.tsx
```diff
@@ -0,0 +1,102 @@
...
+export interface QuicksightDashboardProps
+  extends QuicksightDashboardStateProps,
+    QuicksightDashboardDispatchProps,
+    QuicksightDashboardOwnProps {}
+
+export interface QuicksightDashboardState {
+  embedUrl: string;
+}
+
+class QuicksightDashboard extends React.Component<
```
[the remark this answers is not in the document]
dbernstein: Yes - will rework this once we get some feedback.

src/components/QuicksightDashboard.tsx
```diff
@@ -0,0 +1,102 @@
...
+          "/admin/quicksight_embed/" +
+            this.props.dashboardId +
+            "?libraryUuids=" +
+            library_uuids
+        )
+          .then((response) => response.json())
+          .then((data) => this.setState({ embedUrl: data["embedUrl"] }))
+          .catch((error) => {
+            console.error(error);
+          });
```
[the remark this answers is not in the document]
dbernstein: I changed this to dispatch to an action, but I don't think it is exactly right yet.  I would like to get this out since this is really just a rough pass.

src/components/QuicksightDashboard.tsx
```diff
@@ -0,0 +1,110 @@
...
+    super(props);
+    this.state = { embedUrl: null };
+  }
+  componentDidMount() {
+    if (this.state.embedUrl) {
+      return;
+    }
+    // Every time the component is mounted a fresh embed url must be fetched.
+    // TODO:  For whatever reason, the "this.props.libraries" variable was not being
+    // set when I tried to put this logic in the render method (nor did it work trying to access it in this method).
```
tdilauro, line 53: It doesn't look like `libraries` is being passed as a `prop` from `QuicksightDashboardPage` l.46, so that's probably why we don't see it here. We *are* passing `dashboardId` there, but not `library`.

src/components/QuicksightDashboardPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+    library?: string;
+  };
+}
+
+export interface QuicksightDashboardPageContext {
+  editorStore: Store<RootState>;
+}
+
+/** Page holds quicksight dashboards. */
+export default class DashboardPage extends React.Component<QuicksightDashboardPageProps> {
```
tdilauro, line 21: We should use `QuicksightDashboardPage` -- rather than `DashboardPage` -- here, since it could be confusing. `DashboardPage` is used in the `DashboardPage` component, so using it here could be confusing.

src/index.tsx
```diff
@@ -87,6 +88,10 @@ class CirculationAdmin {
                 path="/admin/web/dashboard(/:library)"
                 component={DashboardPage}
               />
+              <Route
+                path="/admin/web/quicksight(/:library)"
```
tdilauro, line 92: Not sure why we're passing the library here, since it is not currently used to select which `libraryUuid`s are selected for inclusion.

src/components/QuicksightDashboardPage.tsx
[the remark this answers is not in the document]
tdilauro: Since it's a hardcoded value, I'm not sure why we need to pass it as a `prop`. Can we just remove the `prop` here and use the hardcoded value inside the child `QuicksightDashboard` component?

tdilauro, line 46: This is where we can pass the `library` prop into the `QuicksightDashboard` component.

src/components/QuicksightDashboard.tsx
dbernstein: Right - "libraries" is the list of libraries that is being pulled from the store - at least I'd like to be pulling it from the store.  But it doesn't get loaded. I removed the reference to "library" for now, because of the way the QS filtering is working:  it's always returning the list of libraries the user has access to.  So in the library QS view,  users can look across all libraries, any group of libraries or a single library.

src/components/QuicksightDashboardPage.tsx
dbernstein: Good point I'll fix that.

src/index.tsx
dbernstein: Right - I'll remove that for now.   Oversight.

src/components/QuicksightDashboardPage.tsx
dbernstein: For the demo,  we'll leave the library prop off so the user can select the libraries they have access to when viewing the report.

dbernstein:
@tdilauro : updates are in.

src/components/QuicksightDashboardPage.tsx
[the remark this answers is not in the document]
tdilauro: It doesn't look like this has been updated.

src/components/QuicksightDashboardPage.tsx
```diff
@@ -0,0 +1,57 @@
...
+import Footer from "./Footer";
+import title from "../utils/title";
+import QuicksightDashboard from "./QuicksightDashboard";
+
+export interface QuicksightDashboardPageProps
+  extends React.Props<QuicksightDashboardPageProps> {
+  params: {
+    library?: string;
+  };
+}
```
tdilauro, line 15: Since we're no longer passing a `library` param here, we no longer need the `QuicksightDashboardPageProps` interface, so we should get rid of this.

We'll also need to remove it from the class definition below.

src/components/QuicksightDashboardPage.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export interface QuicksightDashboardPageContext {
+  editorStore: Store<RootState>;
+}
+
+/** Page holds quicksight dashboards. */
+export default class QuicksightDashboardPage extends React.Component<
+  QuicksightDashboardPageProps
+> {
```
tdilauro, line 24: Since we no longer need `QuicksightDashboardPageProps`, we can change the props type here to none, like this:
```typescript
export default class QuicksightDashboardPage extends React.Component<never> {
...
```

src/components/QuicksightDashboardPage.tsx
```diff
@@ -0,0 +1,57 @@
...
+    };
+  }
+
+  render(): JSX.Element {
+    const { library } = this.props.params;
+    return (
+      <div className="quicksight-dashboard">
+        <Header />
+        <main className="body">
+          <QuicksightDashboard dashboardId="library" />
```
tdilauro, line 47: Again, we don't need to pass this hardcoded `dashboarId` prop here. We can just use the hardcoded value directly in the child component.

src/components/QuicksightDashboard.tsx
```diff
@@ -0,0 +1,110 @@
...
+  fetchLibraries?: () => Promise<LibrariesData>;
+  fetchQuicksightEmbedUrl?: (
+    dashboardId: string,
+    ld: LibrariesData
+  ) => Promise<QuickSightEmbeddedURLData>;
+}
+
+export interface QuicksightDashboardOwnProps {
+  store?: Store<RootState>;
+  dashboardId?: string;
```
tdilauro, line 27: We won't need `dashboardId` as a prop here, assuming that we stop passing the hardcoded value from the parent page.

src/components/QuicksightDashboard.tsx
```diff
@@ -0,0 +1,110 @@
...
+    // TODO:  For whatever reason, the "this.props.libraries" variable was not being
+    // set when I tried to put this logic in the render method (nor did it work trying to access it in this method).
+    // Clearly I am missing something. Also the embed url should be set via the action/reducer pattern,
+    // but again I wasn't able to get things working following the pattern in Header.tsx so this is where I landed.
+    // It's ugly but it works.
+    // Nevertheless it should be brought into alignment before it goes into production.
+    this.props.fetchLibraries().then((libs) => {
+      try {
+        this.props
+          .fetchQuicksightEmbedUrl(this.props.dashboardId, libs)
```
tdilauro, line 61: Assuming that we stop passing the hardcoded value from the parent page, we should replace `this.props.dashboardId` with a new constant in this module.

src/actions.ts
```diff
@@ -190,6 +190,7 @@ export default class ActionCreator extends BaseActionCreator {
   static readonly RESET_ADOBE_ID = "RESET_ADOBE_ID";
 
   static readonly DIAGNOSTICS = "DIAGNOSTICS";
+  static readonly QUICKSIGHT_EMBEDDED_URL: "QUICKSIGHT_EMBEDDED_URL";
```
tdilauro, line 193: I think I missed this before, but the value of this constant needs to assigned with and equal sign.
```typescript
static readonly QUICKSIGHT_EMBEDDED_URL = "QUICKSIGHT_EMBEDDED_URL";
```

This might be why we were running into trouble using Redux actions dispatch. I also note that the various state constants for the `QUICKSIGHT_EMBEDDED_URL` action are not defined, which would also prevent actions dispatch from working properly. There are assumptions/naming conventions for these on which `fetchJSON` depends. A good template for that would be the `STATS` action:
```typescript
  ...
  static readonly STATS = "STATS";
  ...
  static readonly STATS_REQUEST = "STATS_REQUEST";
  static readonly STATS_SUCCESS = "STATS_SUCCESS";
  static readonly STATS_FAILURE = "STATS_FAILURE";
  static readonly STATS_LOAD = "STATS_LOAD";
  ...
```

tdilauro (commented):
I added a few follow-on comments related to the most recent changes.

Also a reminder about the hardcoded `dashboardId` and follow-on changes that would be needed if it were removed.

src/components/QuicksightDashboard.tsx
dbernstein: We may offer other dashboards down the road depending on the user's permissions.  However for now, I suppose I can make this a constant.

src/components/QuicksightDashboardPage.tsx
dbernstein: We're not passing it now, but we may want to pass it later.  Ie if we want to pre-select the library.  To make this work, we'll need to make some changes on the backend.  But I'm thinking we may want to leave this param in here for the time being.

src/components/QuicksightDashboardPage.tsx
dbernstein: I can remove this - but it was my assumption that the admin interface may provide more than one dashboard depending on the access level of the user.

src/actions.ts
dbernstein: I'll fix.

dbernstein: Good catch.

dbernstein: I'm understanding what you did with the Stats piece now a little better.   I'd like to get this out there for front-end feedback, but I will plan to rework the fetching using RTK before we goes live.

tdilauro (approved):
This looks good for now and doesn't appear to break functionality elsewhere in the app.

PR #100, opened 2024-01-18
Remove search service components (PP-848)

jonathangreen:
## Description

Remove the search service configuration and self test screens from the admin UI.

## Motivation and Context

In PP-93 these components were moved to be configured via environment variables, rather then the admin UI.

## How Has This Been Tested?

Tested locally.

## Checklist:


tdilauro (approved):
Looks good to me! 🔥

PR #106, opened 2024-03-27
Add fiction list filter (PP-1110)

jonathangreen:
## Description

Add a new filter to the list search that allows filtering on `fiction` and `nonfiction`. If the new filter is selected, the input is replaced with a select list, so that the values you can enter are restricted to `fiction` and `nonfiction` since these are the only two valid values for the field. 

Any feedback on this one would be great. It does accomplish what is asked in the ticket, and I think it works from the user perspective, in my testing at least. I only know enough to be dangerous about react though, so there may be better ways to accomplish what I'm doing here.

## Motivation and Context

Would like to be able to filter custom lists by `fiction` status. See PP-1110.

## How Has This Been Tested?

Tested against gorgon running the admin interface locally with the blackstone collection.

## Checklist:

### DIFF
--- a/src/components/AdvancedSearchFilterInput.tsx
+++ b/src/components/AdvancedSearchFilterInput.tsx
@@ -27,11 +27,20 @@ export default function AdvancedSearchFilterInput({
     setFilterKey(value);
     const selected = fields.find((field) => field.value === value);
 
-    // Set the value to either the default option, or blank when changing the filter type
+    // Set the value to either the first option, or blank.
     if (selected.options && selected.options.length) {
       setFilterValue(selected.options[0]);
     } else {
       setFilterValue("");
+    }
+
+    // Update operator if the filter doesn't support it.
+    if (
+      selected.operators &&
+      selected.operators.length &&
+      !selected.operators.find((op) => op === filterOp)
+    ) {
+      setFilterOp(selected.operators[0]);
     }
   };
 
--- a/src/components/AdvancedSearchFilterInput.tsx
+++ b/src/components/AdvancedSearchFilterInput.tsx
@@ -25,6 +25,7 @@ export default function AdvancedSearchFilterInput({
 
   const handleKeyChange = (value) => {
     setFilterKey(value);
+    setFilterValue("");
   };
 
   const handleOpChange = () => {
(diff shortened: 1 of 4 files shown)


src/components/AdvancedSearchFilterInput.tsx
```diff
@@ -64,6 +72,17 @@ export default function AdvancedSearchFilterInput({
   };
 
   const selectedField = fields.find((field) => field.value === filterKey);
+  const selectedFieldOperators =
+    selectedField.operators !== undefined
+      ? operators.filter((element) => {
+          return selectedField.operators.includes(element.value);
+        })
+      : operators;
```
tdilauro, line 80: Minor: This is fine and will work; but we could simplify it. Since there's only one statement in the filter function, we can drop the braces and the `return`. And because there's only one parameter, we can drop the parens. If we don't mind including `null` with the already-present `undefined`, we can also drop the comparison.
```
  const selectedFieldOperators = selectedField.operators
    ? operators.filter(op => selectedField.operators.includes(op.value))
    : operators;
```

src/components/AdvancedSearchFilterInput.tsx
```diff
@@ -25,6 +25,14 @@ export default function AdvancedSearchFilterInput({
...
   const handleKeyChange = (value) => {
     setFilterKey(value);
+    const selected = fields.find((field) => field.value === value);
+
+    // Set the value to either the default option, or blank when changing the filter type
+    if (selected.options && selected.options.length) {
+      setFilterValue(selected.options[0]);
+    } else {
+      setFilterValue("");
+    }
```
tdilauro, line 35: There could be a problem here. This PR is introducing the potential for a reduced set of operators, so we'll also need to ensure that we've got a valid operator for the newly current field. Ideally, we'd only change the `filterOp` if the current one is not supported.

It would be useful to add a test simulating going from an op on a field with broader set ops to a field that doesn't support that op, then adding the filter, and finally verifying that the key, value, and op are as expected.

tdilauro (commented):
This looks pretty good. There is one potential sneaky bug related to restricting `filterOps` when a new `filterKey` is selected.

src/components/AdvancedSearchFilterInput.tsx
jonathangreen: That it much more readable, thanks! `Eslint` doesn't like dropping the braces around `op`, so I added those back in, but otherwise I made this change.

src/components/AdvancedSearchFilterInput.tsx
jonathangreen: Good catch! Thank you, totally missed this one.

jonathangreen:
@tdilauro I think this is ready for another review. I added another test and addressed both your code review comments.

tdilauro (approved):
Looks good! 🎉

PR #112, opened 2024-05-24
Updates `msw` to v2. (PP-1261)

tdilauro:
## Description

- Upgrades `msw` from v1 to v2.
- Polyfills some objects in the `jest` environment to support the change.
- Corrects a handler that was not working as intended, even before the upgrade.

## Motivation and Context

We will be adding more tests that rely on `msw`, so it made sense to upgrade now, rather than having many more tests to upgrade in the not-too-distant future.

[Jira [PP-1261](https://ebce-lyrasis.atlassian.net/browse/PP-1261)]

## How Has This Been Tested?

- All tests pass locally.

## Checklist:

[PP-1261]: https://ebce-lyrasis.atlassian.net/browse/PP-1261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ray-lee (approved):
Looks good!

PR #113, opened 2024-05-29
Introduce React Toolkit.

tdilauro:
## Description

- Introduces React Toolkit.
- Migrates store configuration to RTK.
- Moves most of our Redux imports to RTK.
  - The notable exception is `src/reducers/index.ts` which showed errors when importing `combineReducers` from RTK, rather than "redux".

## Motivation and Context

- Gives us the ability to incrementally simplify and modernize the codebase.

## How Has This Been Tested?

- Some manual testing.
- All automated tests pass.

## Checklist:


src/store.ts
tdilauro, line 1: All of the real changes happen here, where we create a function to configure the store, and in the package files.

All of the rest of the changes are just moving imports.

tdilauro: Here is `src/store.ts`.

jonathangreen (approved):
Didn't get deeply into understanding the change here, but it looks good to me!

PR #123, opened 2024-08-07
Introduce modern context provider and hooks (PP-1572)

tdilauro:
## Description

Adds context hooks for many of the properties provided by the current, legacy context provider in the `ContextProvider` component. With this change, the `ContextProvider` component injects a provider for the new `AppContext` into the component tree, so we don't need to account for another component in the app tree or in test wrappers.

## Motivation and Context

- Provide a migration path away from legacy context provider.
- Make it easier to write some business logic hooks.

[Jira [PP-1572](https://ebce-lyrasis.atlassian.net/browse/PP-1572)]

## How Has This Been Tested?

- Manually tested locally with the dev-server.
- New and existing tests pass locally. Existing tests did not need to be modified.
- [CI tests for associated branch](https://github.com/ThePalaceProject/circulation-admin/actions/runs/10289315684) passed.

## Checklist:

[PP-1572]: https://ebce-lyrasis.atlassian.net/browse/PP-1572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/ContextProvider.tsx
+++ b/src/components/ContextProvider.tsx
@@ -6,6 +6,7 @@ import { FeatureFlags, PathFor } from "../interfaces";
 import Admin from "../models/Admin";
 import PathForProvider from "@thepalaceproject/web-opds-client/lib/components/context/PathForContext";
 import ActionCreator from "../actions";
+import AppContextProvider from "../context/appContext";
 
 export interface ContextProviderProps extends React.Props<ContextProvider> {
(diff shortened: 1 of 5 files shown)


ray-lee (approved):
🚀 This looks like a good way to migrate to modern context.

PR #140, opened 2024-11-22
Fix formatting of long lane names (PP-1962)

jonathangreen:
## Description

In the work editor long lane names were being formatted poorly. Found this while working on PP-1962.

## Motivation and Context

Before this change:
Screenshot 2024-11-22 at 2 30 42 PM

After:
Screenshot 2024-11-22 at 2 29 02 PM

I took a look around the admin interface after making this change, and I don't see any ill effects, but I'm not 100% sure everywhere this style-sheet is used, so its possible there is a better way to approach this.

## How Has This Been Tested?

- Tested locally

## Checklist:

### DIFF
--- a/src/stylesheets/input_list.scss
+++ b/src/stylesheets/input_list.scss
@@ -27,7 +27,6 @@
       display: block;
       background: darken($gray-tint, 5);
       padding: 6px 12px;
-      height: 34px;
       width: 192px;
       border: 0;
     }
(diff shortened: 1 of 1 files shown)


src/stylesheets/input_list.scss
```diff
@@ -27,7 +27,6 @@
       display: block;
       background: darken($gray-tint, 5);
       padding: 6px 12px;
-      height: 34px;
```
tdilauro, line 30: This change is probably safe, but in order to ensure that the element doesn't collapse if it has content that doesn't intrinsically have height of its own, I would suggest changing this property to 
```css
    min-height: 34px;
```
rather than removing height altogether. That will prevent the element collapsing, but will allow it to grow as needed.

tdilauro (approved):
Looks good! One suggestion below.

PR #141, opened 2024-11-22
Remove max_cache_age parameter (PP-1962)

jonathangreen:
## Description

The code was doing some questionable things to add the `max_cache_age` parameter to URLs. In PP-1962 I found this is causing some issues. We could do some work to fix up how we include the parameter in URLs, but since we are no longer using this parameter in the CM, it doesn't seem worth including it.

## Motivation and Context

Issues I was seeing with this code: 
- max_cache_age would get repeatedly included creating urls like this: `http://localhost:6500/admin/web/collection/GA0011%3Fmax_cache_age%3D0%26max_cache_age%3D0%26max_cache_age%3D0%26max_cache_age%3D0/book/GA0011%2FISBN%2F9780593457580/tab/lists` 🤮 
- Clicking the **Create New List** button in the admin UI would go to a URL like this `http://localhost:6500/admin/web/lists/GA0011?max_cache_age=0&max_cache_age=0&max_cache_age=0&max_cache_age=0/create` which isn't a routable URL so it resulted in a white screen. The URL we wanted was `http://localhost:6500/admin/web/lists/GA0011/create`.
- I saw a couple other weird URL generation quirks while looking at this that I didn't note down. But removing the `max_cache_age` stuff fixed them

It seems like we might want to do some refactoring work to how the admin UI is generating URLs in the future, but this at least solves the issues I was seeing with the work lane editor screens.

## How Has This Been Tested?

- Tested locally
- Running unit tests

## Checklist:

### DIFF
--- a/src/components/CatalogPage.tsx
+++ b/src/components/CatalogPage.tsx
@@ -103,11 +103,7 @@ export default class CatalogPage extends React.Component<CatalogPageProps, {}> {
   }
 
   expandCollectionUrl(url: string): string {
-    return url
-      ? `${document.location.origin}/${url}${
-          url.includes("?") ? "&" : "?"
-        }max_cache_age=0`
-      : url;
+    return url ? `${document.location.origin}/${url}` : url;
   }
 
   expandBookUrl(url: string): string {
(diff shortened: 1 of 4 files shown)


tdilauro (approved):
Nice! No más max_cache! 🎉

PR #142, opened 2024-11-22
Increase WebPack dev-server proxy timeout.

tdilauro:
## Description

Increase the `dev-server` proxy timeout to 120 seconds.

## Motivation and Context

Sometimes the proxy server would time out, especially when performing a catalog request, but sometimes during other functions. These timeouts are not helpful, so bumping up to a high value should eliminate them.

## How Has This Been Tested?

- Tested locally against a library that was consistently causing timeouts.
- This change does not impact any tests in the test suite.

## Checklist:

- N/A - All new and existing tests passed.

### DIFF
--- a/webpack.dev-server.config.js
+++ b/webpack.dev-server.config.js
@@ -192,11 +192,11 @@ module.exports = (env) => {
         return responseBuffer;
       }
     ),
-    proxyTimeout: 5000,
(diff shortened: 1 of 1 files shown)


jonathangreen (commented):
🚀

jonathangreen (approved):
🚀 mark 2

PR #152, opened 2025-04-16
No cache for catalog fetch. (PP-2398)

tdilauro:
## Description

Uses a custom opds-web-client data fetcher in the `CatalogPage` component to fetch the catalog with no caching.

## Motivation and Context

We need to bypass the cache in the Admin UI so that we can see the results of our changes immediately.

[Jira PP-2398]

## How Has This Been Tested?

- No tests for this yet.
- Manual testing fails because there is a bug in the opds-web-client `Root` HOC
  - In the Root component, `mergeRootProps` is imported and [used in the creation of the exported HOC from the class](https://github.com/thepalaceproject/web-opds-client/blob/c216483c6c0f02ba17696ff99314cff45bc1a548/packages/web-opds-client/src/components/Root.tsx#L454-L462). Unfortunately, [it creates its own fetcher](https://github.com/thepalaceproject/web-opds-client/blob/c216483c6c0f02ba17696ff99314cff45bc1a548/packages/web-opds-client/src/components/mergeRootProps.ts#L144-L152), so the one we supply is not used.

## Checklist:


src/CustomHeaderDataFetcher.ts
```diff
@@ -0,0 +1,24 @@
...
+    extraHeaders = {},
+    ...dataFetcherConfig
+  }: CustomHeaderDataFetcherConfig = {}) {
+    super(dataFetcherConfig);
+    this.extraHeaders = extraHeaders;
+  }
+
+  fetch(url: string, { headers = {}, ...restOptions } = {}) {
+    const customizedHeaders = Object.assign({}, this.extraHeaders, headers);
+    return super.fetch(url, { headers: customizedHeaders, ...restOptions });
```
jonathangreen, line 22: I'm not 100% on this, but I was wondering if we should modify this to be `NoCacheDataFetcher`, and make sure we are bypassing the browser cache as well by setting the `cache: no-cache` option to fetch as well.

I'm not totally clear if we also need to set the headers when setting the cache option, or if that option will set the headers as well.

tdilauro: Sure. We can do that. I'm happy to make those changes, if you like. Otherwise, feel free to take this PR wherever you want. This was mainly to see what would be necessary. And it provides a bit of roadmap of how to do it and type it without pulling in much from the superclass.

jonathangreen:
@tdilauro I'm going to close this one, and put up a PR with my changes, so you can review it.

PR #153, opened 2025-04-16
Fix timezone parsing in tests

jonathangreen:
## Description

  Fixed failing tests related to time zone handling in SelfTests component tests. Updated the test expectations to dynamically compute expected date/time strings based on the actual date object from the test data, rather than hardcoding specific time values.

##  Motivation and Context

  Previously, the SelfTests component tests were failing because they expected a specific hardcoded time output ("Tests last ran on Tue Aug 07 2018 15:34:54 and lasted 1.75s.") without accounting for the developer's local timezone. This caused tests to fail depending on the timezone of the developer's machine.

The admin UI outputs the localized time, which is the correct behavior, but since the tests were not adjusting for timezones, it meant that anyone running the tests whose local timezone wasn't US eastern would get a test failure.

These tests have been failing on my machine for a long time, finally decided to fix it.

## How Has This Been Tested?

  Ran the test suite locally to verify that the previously failing tests now pass. The fix maintains the integrity of the test expectations while making them work in any timezone.

## Checklist

### DIFF
--- a/src/components/__tests__/SelfTests-test.tsx
+++ b/src/components/__tests__/SelfTests-test.tsx
@@ -182,8 +182,15 @@ describe("SelfTests", () => {
   });
 
   it("should format the date and duration of the most recent tests", () => {
+    // Create a date object from the ISO string to account for local timezone
+    const date = new Date(collections[0].self_test_results.start);
+    const expectedDate = date.toDateString();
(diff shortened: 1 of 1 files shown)


tdilauro (approved):
Looks good!

PR #159, opened 2025-05-30
Correctly reset state on manual custom list detail load. (PP-2447, PP-2076)

tdilauro:
## Description

Correctly reset locally added / removed entries when reloading a manual custom list after save, since the associated changes should already be reflected in fresh state just synced from the server. 

## Motivation and Context

- Attempting more than one save of a manual custom list without reloading the page between saves would result in all attempts after the first one to fail.
- If items were added, they would appear both as unsaved additions (indented at the top of the list) and lower down, unindented in the saved items list.

[Jira PP-2447, PP-2076]

## How Has This Been Tested?

- Manually testing via dev-server against a staging instance.
- All tests pass locally.
- [CI tests](https://github.com/ThePalaceProject/circulation-admin/actions/runs/15357194217) pass.

## Checklist:

### DIFF
--- a/src/components/AdvancedSearchBuilder.tsx
+++ b/src/components/AdvancedSearchBuilder.tsx
@@ -52,9 +52,7 @@ export const fields: FieldType[] = [
     operators: ["eq"],
   },
 ]
-  // This is a work-around to get the functionality of `Array.toSorted`.
-  .slice()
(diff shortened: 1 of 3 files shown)


jonathangreen (approved):
Looks good!

PR #164, opened 2025-06-30
Upgrade `webpack-dev-server` from v4 to v5. (PP-2687)

tdilauro:
## Description

- Upgrades `webpack-dev-server` from v4 to v5.
- Update dev server proxy code to support new v5 api.

## Motivation and Context

- [Dependabot wanted to upgrade `webpack-dev-server` from v4 to v5](https://github.com/ThePalaceProject/circulation-admin/pull/160); however, when I did some manual compatibility checking, I saw that the proxy server would no longer work. So, the upgrade and the proxy code update had to be coordinated, which Dependabot was not doing

## How Has This Been Tested?

Manual testing in local development environment.

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -110,7 +110,7 @@
         "url-loader": "^4.1.1",
         "webpack": "^5.94.0",
         "webpack-cli": "^5.0.1",
-        "webpack-dev-server": "^4.11.1",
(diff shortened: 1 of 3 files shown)


PoorBillyPilgrim (approved):
LGTM 👍

PR #213, opened 2026-04-13
Show / hide configuration associations (PP-4022)

tdilauro:
## Description

Add an expandable disclosure panel to each row on the integration/configuration list pages, showing the libraries / roles associated with that service. Users can expand/collapse individual rows or use Expand all / Collapse all controls. `Alt+Click` on any toggle expands or collapses all rows at once.

This is implemented as a set of overridable template methods on `GenericEditableConfigList`:

- `getAssociatedEntries` — returns entries to display (or `undefined` to suppress the toggle entirely)
- `formatAssociatedCount` — controls the count label (e.g. "3 libraries", "2 roles", "1 registered library")
- `getAllLibraries` — resolves short names to display names and edit links

Each component overrides as needed:
- **Collections / base class** — shows associated libraries
- **DiscoveryServices** — shows successfully registered libraries with their registration stage
- **IndividualAdmins** — shows admin roles; sitewide roles (manager-all, librarian-all) are pinned to the top; system admins show a synthetic "sysadmin" entry
- **Libraries** — suppresses the panel (libraries are not associated with other libraries)

Also includes a few minor fixes picked up along the way:
- `UNSAFE_componentWillMount` update to `componentDidMount`, 
- `(window as any).FormData` updated to `FormData`, 
- `.catch()` added to `save()` to suppress unhandled-rejection warnings, and 
- `CollectionData` now explicitly declares `marked_for_deletion`.

## Motivation and Context

Admins currently have no way to see which libraries are associated with a service without opening each service's edit form individually.

[Jira PP-4022]

## How Has This Been Tested?

- Manual testing in local development environment.
- New tests for added functionality.
- [CI checks](https://github.com/ThePalaceProject/circulation-admin/actions/runs/24351053998) pass.

## Checklist:

### DIFF
--- a/src/components/EditableConfigList.tsx
+++ b/src/components/EditableConfigList.tsx
@@ -10,6 +10,7 @@ import ErrorMessage from "./ErrorMessage";
 import PencilIcon from "./icons/PencilIcon";
 import TrashIcon from "./icons/TrashIcon";
 import VisibleIcon from "./icons/VisibleIcon";
+import DisclosureIcon from "./icons/DisclosureIcon";
 import Admin from "../models/Admin";
 import * as PropTypes from "prop-types";
 
@@ -76,11 +77,15 @@ export interface ExtraFormSectionProps<T, U> {
 
     GenericEditableConfigList allows subclasses to define additional props. Subclasses of
     EditableConfigList cannot change the props and do not have to specify a type for them. */
+interface EditableConfigListState {
+  expandedItems: Record<string, boolean>;
+}
+
 export abstract class GenericEditableConfigList<
   T,
   U,
   V extends EditableConfigListProps<T>
-> extends React.Component<V> {
+> extends React.Component<V, EditableConfigListState> {
   context: { admin: Admin };
   static contextTypes = {
     admin: PropTypes.object.isRequired,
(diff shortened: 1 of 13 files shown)


src/components/EditableConfigList.tsx
```diff
@@ -165,11 +196,18 @@ export abstract class GenericEditableConfigList<
...
-                this.renderLi(item, index)
+              {itemsWithEntries.map(([item, entries]) =>
+                this.renderLi(item, entries)
               )}
             </ul>
+            {/* aria-hidden: the bottom controls are a visual convenience
+                duplicate of the controls above the list. Screen readers and
+                keyboard users should interact with the first set only. */}
+            <div aria-hidden="true">
+              {this.renderExpandCollapseControls(expandable)}
```
jonathangreen: Minor comment / question on this one: Does `aria-hidden` mean that keyboard users can't land on this control? Otherwise couldn't a user tab into this control, but not get the announcement for it?

src/components/__tests__/EditableConfigList-test.tsx
```diff
@@ -596,4 +596,81 @@ describe("EditableConfigList", () => {
     wrapper.setContext({ admin: librarian });
     expect(wrapper.instance().getAdminLevel()).to.equal(1);
   });
+
+  describe("renderAssociatedLibraries", () => {
```
jonathangreen: Minor: Since we add `tests/jest/components/EditableConfigList.test.tsx`, shouldn't these new tests go there instead?

jonathangreen (approved):
This looks great to me @tdilauro! This is a feature that I've wanted to see for a long time. 🎉

As always, my caveat with front-end reviews is that I only know enough here to be dangerous. I added a couple of minor comments to consider, but they may or may not be relevant.

src/components/EditableConfigList.tsx
tdilauro: Good catch. I forgot to take it out of the tab order.

src/components/__tests__/EditableConfigList-test.tsx
tdilauro: I put these in here because I was being lazy and using the old enzyme style tests. But I'll move them over to jest/react testing library.

PR #224, opened 2026-05-24
Adds JSON form field support (PP-4438)

tdilauro:
## Description

Adds a new `JsonField` form component for editing JSON-typed protocol settings. The field renders as a `textarea` with:
- Live JSON validation with an inline error message;
- Copy-to-clipboard and clear buttons that appear on hover/focus;
- One-shot undo (Ctrl-Z / Cmd-Z) after a clear;
- Accessible ARIA markup.

**Note**: This field is activated by the new `FormFieldType.JSON` settings field type introduced on [circulation PR #3386](https://github.com/ThePalaceProject/circulation/pull/3386).

## Motivation and Context

Allows admins to actively validate JSON in place without submitting the form.

[Jira PP-4438]

## How Has This Been Tested?

- Manual testing in local development environment.
- Additional tests to cover the new functionality.
- All checks pass locally.
- [CI checks](https://github.com/ThePalaceProject/circulation-admin/actions/runs/26348949842) pass.

## Checklist:

### DIFF
--- a/src/components/JsonField.tsx
+++ b/src/components/JsonField.tsx
@@ -1,0 +1,105 @@
+import * as React from "react";
+import { SettingData } from "../interfaces";
+
+export interface JsonFieldProps {
+  setting: SettingData;
+  value?: any;
+  disabled?: boolean;
+  readOnly?: boolean;
+  onChange?: (value: any) => void;
+}
+
+export interface JsonFieldState {
+  text: string;
+  jsonError: string | null;
+}
+
+function valueToText(value: any): string {
+  if (value === null || value === undefined) return "";
+  return JSON.stringify(value, null, 2);
+}
+
+export default class JsonField extends React.Component<
+  JsonFieldProps,
+  JsonFieldState
+> {
+  static displayName = "JsonField";
+
+  constructor(props: JsonFieldProps) {
+    super(props);
+    this.state = {
+      text: valueToText(props.value),
+      jsonError: null,
+    };
+    this.handleChange = this.handleChange.bind(this);
+  }
+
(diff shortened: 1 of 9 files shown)


tdilauro:
> **Claude finished @tdilauro's task in 4m 34s** —— [View job](https://github.com/ThePalaceProject/circulation-admin/actions/runs/26350287085)
> 
> ...
> 
> ### Details
> #### `src/components/JsonField.tsx:135`
> `handleCopy` calls `navigator.clipboard.writeText(text)` directly without first checking whether `navigator.clipboard` exists. The Clipboard API is only available in secure contexts (HTTPS); in an HTTP dev environment, `navigator.clipboard` is `undefined` and this line throws a `TypeError` that escapes the React event handler, since React 16 does not route event-handler exceptions through error boundaries. A guard prevents the crash:
> 
> ```tsx
> function handleCopy() {
>   if (!navigator.clipboard) return;
>   if (copyTimeoutId.current) clearTimeout(copyTimeoutId.current);
>   navigator.clipboard.writeText(text).then(...);
> }
> ```
> 
> https://github.com/ThePalaceProject/circulation-admin/blob/e8f5eb585ab1cf22f6fce1cc0a9e4ec98df7da92/src/components/JsonField.tsx#L135
> 
> ...

Fair point. However, the suggested change will not work. Or, more to the point, it won't produce the desired result, which would be to display the "Copy failed" message so that the user knows that something went wrong. It will just fail silently.

I fixed this and actually made the code a little more readable in the process. Also addressed the other minor / nit issues raised in the initial AI review.

This is ready for human review now.

jonathangreen (approved):
Looks great @tdilauro!

PR #254, opened 2026-06-08
Migrate to ESLint 9 with flat config

jonathangreen:
## Description

Upgrades ESLint from 7 to **9** and migrates from the legacy `.eslintrc` to the **flat config** format (`eslint.config.mjs`).

Dependency changes (relative to `main`):
- `eslint` `^7.0.0` → `^9.39.4`; add `@eslint/js` `^9.39.4` and `globals` `^16.5.0`
- replace `@typescript-eslint/eslint-plugin` + `@typescript-eslint/parser` (v5) with the unified `typescript-eslint` `^8.60.1`
- `eslint-config-prettier` `^6.11` → `^10.1.8`
- `eslint-plugin-jsx-a11y` `^6.2.3` → `^6.10.2`
- `eslint-plugin-react` `^7.19.0` → `^7.37.5`
- **remove** `eslint-plugin-prettier` — Prettier is kept at 2.x and still runs standalone in the lint-staged hook; `eslint-config-prettier` continues to silence conflicting rules. (Keeping the integration would have forced a Prettier 3 upgrade and a codebase-wide reformat.)
- `eslint-plugin-react-hooks` is **unchanged** here — `main` is already at `^7.1.1`. This PR wires the plugin into the flat config (it was previously an installed-but-unused dependency).

### Why ESLint 9 and not 10?

ESLint 10 was deferred because `eslint-plugin-react` and `eslint-plugin-jsx-a11y` have not yet released versions that declare ESLint 10 peer support (their latest published releases predate ESLint 10's Feb 2026 stable). Both have active "ESLint v10 compatibility" tracking issues, so this is a timing gap, not abandonment. ESLint 9 installs cleanly with **no peer overrides**, and makes the eventual bump to 10 a one-line change once those plugins ship compatible releases.

## Motivation and Context

Dependabot opened #250 to bump ESLint 7 → 10, which failed CI on an `npm ci` `ERESOLVE` peer-dependency conflict: the project's toolchain (`@typescript-eslint` v5, etc.) is incompatible with modern ESLint. That major bump was closed/ignored in favor of this deliberate migration onto a supported, flat-config toolchain.

## Follow-up PRs (stacked on this one)

This is the base of a three-PR stack. Merge in order:

1. **This PR** — ESLint 9 + flat config migration (parity; known latent findings left in place).
2. **#255 — Fix latent code issues surfaced by ESLint** — resolves the `rules-of-hooks` and `display-name` errors and the intentional `exhaustive-deps` effects that this migration surfaces.
3. **#256 — Clear ESLint warnings and enforce ESLint in CI** — clears the ~187 `no-unused-vars` warnings and the stale `eslint-disable` directives, then adds a CI gate (`eslint . --max-warnings 0`) so nothing new sneaks in.

## How Has This Been Tested?

- `npm install` resolves cleanly — no `ERESOLVE`, no peer overrides required.
- Full test suite passes (`npm test`): mocha **1062 passing**, jest **349 passed / 32 suites**.
- Verified ESLint runs across the repo and that the lint-staged `eslint --cache --fix` pre-commit path works under flat config.

## Checklist:


eslint.config.mjs
```diff
@@ -0,0 +1,97 @@
...
+      "no-useless-escape": 0,
+      "@typescript-eslint/explicit-function-return-type": 0,
+      "react/prop-types": 0,
+      "react/no-string-refs": 0,
+      "jsx-a11y/label-has-associated-control": 0,
+      "react/no-find-dom-node": 0,
+      "react/no-unescaped-entities": 0,
+      "@typescript-eslint/no-inferrable-types": 0,
+      // The following three rules are newly enabled by typescript-eslint v8's
+      // `recommended` preset but were not active under the previous v5 config.
```
tdilauro: Minor - I generally prefer not to reference previous releases in the code base since the comment becomes stale, but it would be a helpful PR comment for these lines. I think the rest of the comment (from L60) explains the situation pretty well on it's own.


```suggestion
```

tdilauro (approved):
Looks good!

One minor comment about comments.

PR #255, opened 2026-06-08
Fix latent code issues surfaced by ESLint

jonathangreen:
## Description

Stacked on #254 (base branch `eslint-9-migration`). Fixes the real code issues that the ESLint 9 migration surfaced once `eslint .` lints the whole tree (the pre-commit hook only lints *staged* files, so these latent issues had accumulated unseen).

- **`AdvancedSearchBooleanFilter.tsx`** — `react-hooks/rules-of-hooks`: `useDrop` was called after an early `return null`. Moved the `!children` guard to *after* the hook so hooks always run in the same order. `query` is non-null in that guard, so behavior is unchanged.
- **`AdvancedSearchValueFilter.tsx`** — `react/display-name`: it was an anonymous `export default ({...}) => {...}`. Named the component. Because naming it makes `react-hooks` analyze it, that surfaced the *same* latent rules-of-hooks pattern (an early `!query` return before `useDrag`/`useDrop`), so the guard was moved below the hooks and the hooks' `query` access made null-safe (`query?.id`) to preserve the defensive behavior.
- **`CustomListEditor.tsx`, `CustomListSearch.tsx`, `JsonField.tsx`** — `react-hooks/exhaustive-deps`: suppressed on three intentional effects (a load-triggered search, a mount-only query seed, and a reusable timeout-cleanup hook), each with a comment explaining why adding the missing deps would be incorrect (re-trigger loops / duplicate searches).

## Motivation and Context

#254 deliberately left these latent findings out of scope to keep the dependency migration clean. This PR resolves them so the tree can reach a zero-error baseline ahead of enabling a full ESLint gate in CI (next stacked PR).

## How Has This Been Tested?

- `eslint .` reports **0 errors and 0 `exhaustive-deps` warnings** after the change (only the pre-existing `no-unused-vars` warnings remain, handled in the follow-up).
- Full `npm test` passes: mocha **1062 passing**, jest **349 passed / 32 suites**. The four affected components have existing test coverage under `src/components/__tests__/`.

## Checklist:

### DIFF
--- a/src/components/AdvancedSearchBooleanFilter.tsx
+++ b/src/components/AdvancedSearchBooleanFilter.tsx
@@ -33,10 +33,6 @@ export default function AdvancedSearchBooleanFilter({
 }: AdvancedSearchBooleanFilterProps) {
   const children = query && (query.and || query.or);
 
-  if (!children) {
-    return null;
-  }
-
   const handleBooleanChange = (
     event: React.SyntheticEvent<HTMLSelectElement>
   ) => {
@@ -93,6 +89,11 @@ export default function AdvancedSearchBooleanFilter({
     },
     [query.id, onMove]
   );
+
+  // Hooks must run unconditionally, so this guard comes after useDrop above.
+  if (!children) {
+    return null;
+  }
 
   const isSelected = !readOnly && selectedQueryId === query.id;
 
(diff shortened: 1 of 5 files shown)


tdilauro (approved):
🚀

PR #273, opened 2026-06-11
Add a Queue Reap button to the collection editor (PP-4363)

jonathangreen:
## Description

Adds a "Queue Reap" button to the collection editor. Specifically:

- Adds a `REAP_COLLECTION` action that posts to `/admin/collection/<id>/reap`, a `CollectionReapButton` gated on the protocol's `supports_reap` flag, and wires `reapCollection` through the `Collections` context and dispatch.
- Extracts the import button's in-flight/feedback state into a shared `useCollectionTask` hook and its markup into a shared `CollectionTaskPanel`, so import and reap stay in sync.
- Styles the reap button and the forced full re-import button as heavyweight tasks via a shared `collection-task-caution` class (yellow, cautionary copy).

## Motivation and Context

Reaping for collections only happens on a cron schedule. When items are accidentally added to a collection's feed and show up in the catalog, there is currently no way to remove them without filing a request for a manual reap. This gives admins a self-service control to trigger a reap immediately.

This is the frontend half of the change. It depends on the backend endpoint and `supports_reap` protocol flag added in ThePalaceProject/circulation#3454, and the two must be deployed together.

## How Has This Been Tested?

- Added `CollectionReapButton.test.tsx` (renders only when `supports_reap` is set, hidden for unsaved collections and older CMs that omit the flag, triggers reap with the collection id, success/error feedback, disabled/loading states) and updated the existing Collections tests for the new prop.
- Ran the affected Jest suites (passing) and lint (`eslint --max-warnings 0` + sass-lint, clean).

## Checklist:

### DIFF
--- a/src/components/CollectionTaskPanel.tsx
+++ b/src/components/CollectionTaskPanel.tsx
@@ -1,0 +1,58 @@
+import * as React from "react";
+import { Panel } from "library-simplified-reusable-components";
+
+export interface CollectionTaskPanelProps {
+  /** DOM id for the underlying Panel. */
+  id: string;
+  /** Collapsible panel header text. */
+  headerText: string;
+  /**
+   * The edited collection's id. Used as the `<details>` key so the
+   * "More details" disclosure collapses when the collection changes.
+   */
+  collectionId?: string | number;
+  /** The success or error message to display, or null when idle. */
+  feedback: string | null;
+  /** True when `feedback` describes a success rather than a failure. */
+  success: boolean;
+  /** The action button(s) and any related controls. */
+  controls: React.ReactNode;
+  /** Short inline description shown under the controls. */
+  description: React.ReactNode;
+  /** Long-form documentation revealed under "More details". */
+  docs: React.ReactNode;
+}
+
+/**
+ * Shared panel layout for collection task buttons (import, reap, ...).
+ * Renders the supplied controls, feedback, a short description, and a
+ * collapsible "More details" docs section inside a Panel.
+ */
+const CollectionTaskPanel: React.FC<CollectionTaskPanelProps> = ({
+  id,
+  headerText,
+  collectionId,
+  feedback,
+  success,
+  controls,
+  description,
+  docs,
+}) => {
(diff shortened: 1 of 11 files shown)


src/components/CollectionTaskPanel.tsx
```diff
@@ -0,0 +1,58 @@
...
+  controls,
+  description,
+  docs,
+}) => {
+  const feedbackClass = success ? "alert alert-success" : "alert alert-danger";
+
+  const content = (
+    <div className="collection-task">
+      {controls}
+      {feedback && <div className={feedbackClass}>{feedback}</div>}
```
tdilauro: We need a `role` or `aria-live` prop here so that ARIA will try to activate this `div` for screen readers when it's injected. I just recently needed to do this for `src/components/JsonField.tsx`

```suggestion
  const feedbackClass = success ? "alert alert-success" : "alert alert-danger";
  const feedbackAriaLive = success ? "polite" : "assertive";

  const content = (
    <div className="collection-task">
      {controls}
      {feedback && (
        <div className={feedbackClass} aria-live={feedbackAriaLive}>
          {feedback}
        </div>
      )}
```

src/stylesheets/collection.scss
```diff
@@ -64,12 +52,29 @@
...
+    background: $yellow;
+    border-color: darken($yellow, 25%);
+    color: $dark-gray;
   }
+}
+
+.collection-import-controls {
+  display: flex;
+  align-items: center;
+  gap: 1em;
```
tdilauro: It looks like `collection-import-controls` was pulled out of what is now `collection-task`. Was that intentional?

tdilauro (approved):
This looks great! ☠️☠️☠️

One a11y suggestion and a CSS question below.

src/components/CollectionTaskPanel.tsx
jonathangreen: Done

src/stylesheets/collection.scss
jonathangreen: Good catch, this wasn't intentional. Fixed!

PR #279, opened 2026-06-18
Update web-opds-client to v1.2.0

jonathangreen:
## Description

Bumps `@thepalaceproject/web-opds-client` from `^1.1.0` to `^1.2.0`.

The dependency changes are confined to web-opds-client and its nested tree. Notably, v1.2.0 brings in an updated nested `jsdom`/`data-urls`, where `@exodus/bytes` declares an *optional* peer dependency on `@noble/hashes` (`^1.8.0 || ^2.0.0`).

## Motivation and Context

[web-opds-client v1.2.0](https://github.com/ThePalaceProject/web-opds-client/releases/tag/v1.2.0) fixes [web-opds-client#111](https://github.com/ThePalaceProject/web-opds-client/pull/111): OPDS feeds are now classified using the server's declared `kind` in the response `Content-Type` header rather than guessed from feed structure. This corrects feeds from libraries without patron authentication (which omit borrow links) being misclassified as navigation feeds — so books render as cover images in lanes instead of gray title boxes.

`NavigationFeed` and `AcquisitionFeed` share the same shape, so the fix is internal to web-opds-client with no breaking API changes and no consuming-side code changes required here.

## How Has This Been Tested?

- `npm ci` — succeeds, lockfile in sync (exit 0)
- `npm run lint:js` — clean
- `npm test` — mocha legacy suite + jest: 33 suites, 363 tests passing

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
         "@nypl/dgx-svg-icons": "0.3.14",
         "@reduxjs/toolkit": "^2.2.5",
         "@tanstack/react-query": "^4.36.1",
-        "@thepalaceproject/web-opds-client": "^1.1.0",
+        "@thepalaceproject/web-opds-client": "^1.2.0",
         "bootstrap": "^3.3.6",
         "classnames": "^2.3.1",
         "draft-convert": "^2.1.5",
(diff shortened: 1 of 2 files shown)


jonathangreen:
I think Claude's comment about `opds-feed-parser@0.0.17` is reasonable, but I'll handle it in a separate PR: https://github.com/ThePalaceProject/circulation-admin/pull/280

tdilauro (approved):
👍🏽

PR #280, opened 2026-06-18
Align direct opds-feed-parser dependency to 0.1.0

jonathangreen:
> **Stacked on #279.** Its base is the `chore/update-web-opds-client-1.2.0` branch and the diff shown here is just the `opds-feed-parser` change. GitHub will auto-retarget this to `main` once #279 merges. It should merge after #279.

## Description

Bumps the direct `opds-feed-parser` dependency from `0.0.17` to `0.1.0` to match the version web-opds-client 1.2.0 uses internally, and updates the two `editorAdapter` mocha fixtures for 0.1.0's stricter `OPDSEntryArgs`.

Lockfile effect (net −41 lines): the tree collapses to a single `opds-feed-parser@0.1.0` / `xml2js@0.6.2` copy instead of two, and the deprecated `core-js@2.6.12` and `requirejs` transitive deps (pulled by 0.0.17) are dropped.

## Motivation and Context

web-opds-client 1.2.0's `DataFetcher` parses feeds with `opds-feed-parser@0.1.0` and hands the resulting `OPDSEntry` to `editorAdapter` (`new DataFetcher({ adapter: editorAdapter })`). Until now `editorAdapter`'s types — and the legacy mocha fixtures in `editorAdapter-test.ts` that call `new OPDSEntry(...)` — came from the directly-declared `0.0.17`, so in production the entries were produced by `0.1.0` while the types came from `0.0.17`. The build still compiled (structural typing) and the `data.unparsed[...]` reads are try/catch-guarded, so this was not a proven break, but bumping the direct dependency keeps the type definitions and fixtures aligned with the parser that actually produces the entries and avoids carrying two copies of `opds-feed-parser`.

`opds-feed-parser@0.1.0` redefines `OPDSEntryArgs extends OPDSEntry`, making every `OPDSEntry` field required — `subtitle` and `series` were optional in `0.0.17`. The two visibility-status fixtures omitted them, so they now provide both (matching the existing "valid OPDS entry" fixture). `editorAdapter.ts` itself needed no changes.

## How Has This Been Tested?

Validated on the same toolchain CI uses — **Node 20 / npm 10.8.2**:

- `npm ci` — lockfile in sync (exit 0)
- `npx tsc --noEmit` — typechecks clean against 0.1.0
- `npm run lint:js` — clean
- `npm run test-file-ts src/__tests__/editorAdapter-test.ts` — 4 passing against 0.1.0
- `npm test` — full mocha + jest suite: 33 suites, 363 tests passing

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
         "@nypl/dgx-svg-icons": "0.3.14",
         "@reduxjs/toolkit": "^2.2.5",
         "@tanstack/react-query": "^4.36.1",
-        "@thepalaceproject/web-opds-client": "^1.1.0",
+        "@thepalaceproject/web-opds-client": "^1.2.0",
         "bootstrap": "^3.3.6",
         "classnames": "^2.3.1",
         "draft-convert": "^2.1.5",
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
👍🏽

PR #283, opened 2026-06-18
Make lane count optional and stop displaying it (PP-4506)

dbernstein:
## Description

The Palace Manager backend is removing lane-size maintenance (see ThePalaceProject/circulation#3424), so the admin `GET .../lanes` response no longer includes a per-lane `count`.

- `interfaces.ts`: make `LaneData.count` optional (`count?: number`) for backward compatibility with older servers.
- `Lane.tsx`: stop rendering the `" (count)"` suffix next to the lane name.

`count` was display-only — no logic depended on it — so this degrades gracefully and is safe to ship independently of the backend change.

## Motivation and Context

JIRA: PP-4506

Companion to ThePalaceProject/circulation#3424, which removes the lane `size`/`count` from the backend.

## How Has This Been Tested?

- `tsc` type-check passes.
- `tests/jest/components/Lane.test.tsx` passes (2/2); Prettier clean.

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -511,7 +511,10 @@ export interface LaneData {
   id: string | number;
   display_name: string;
   visible: boolean;
-  count: number;
+  // Deprecated: the backend no longer returns a lane count (lane sizes were
+  // removed). Kept optional for backward compatibility with older servers; the
+  // UI no longer displays it.
+  count?: number;
   sublanes: LaneData[];
   custom_list_ids: number[];
   inherit_parent_restrictions: boolean;
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -514,6 +514,9 @@ export interface LaneData {
   // Deprecated: the backend no longer returns a lane count (lane sizes were
   // removed). Kept optional for backward compatibility with older servers; the
   // UI no longer displays it.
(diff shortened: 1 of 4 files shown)


src/interfaces.ts
```diff
@@ -511,7 +511,10 @@ export interface LaneData {
   id: string | number;
   display_name: string;
   visible: boolean;
-  count: number;
+  // Deprecated: the backend no longer returns a lane count (lane sizes were
+  // removed). Kept optional for backward compatibility with older servers; the
+  // UI no longer displays it.
+  count?: number;
```
tdilauro, line 520: Minor: Add a `TODO:` about when / under what circumstances this optional and the deprecation comment can be removed.

tdilauro (approved):
Looks good! 🔥

One comment below for adding a TODO for future clean up.

src/interfaces.ts
dbernstein: Done in 700b8af80 — added a TODO noting that `count` and the deprecation comment can be removed once the backend changes in ThePalaceProject/circulation#3424 have been released to production.

dbernstein: I also create a follow up [JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-4648).

PR #296, opened 2026-06-22
Test multiple node versions (PP-4612)

tdilauro:
## Description

- Expands the CI test matrix to run against Node 20, 22, and 24 (previously only Node 20).
- Pins npm to 11.17.0 across all CI workflows (`test`, `publish`, `gh-pages`) so every job uses the same npm version regardless of what ships with the runner's Node installation.
- Updates `packageManager` in `package.json` to `npm@11.17.0` so Dependabot generates lockfiles with the same npm version, preventing the lockfile drift that caused `npm ci` failures.
- Raises `engines.npm` to `>=11.10.0`, which is the minimum version with npm's supply-chain attack mitigations (e.g., `min-release-age`).
- Pins husky to `4.3.8` (the last v4 release) to preserve the existing `package.json`-based hook configuration without requiring a migration to Husky v5+'s shell-script format.
- Removes the now-redundant "Upgrade npm for OIDC support" step from `publish.yml`; the global pin already satisfies that requirement.
- Regenerates the lockfile with npm 11.17.0.
- Fixes a regression we introduced by limiting the npm version to `<11.0.0`.

## Motivation and Context

Node 22 is current LTS and Node 24 is the upcoming LTS. Testing only on Node 20 left gaps where breakage on newer runtimes would go undetected.

[Jira PP-4612]

## How Has This Been Tested?

- The test suite (`npm test`) ran as part of CI on all three Node versions. [The matrix job](https://github.com/ThePalaceProject/circulation-admin/actions/runs/27976386674) itself is the validation. It was green for all three version -- 20, 22, and 24 -- so the testing change is verified.
- We will still need to test a release to verify that the publishing regression is resolved.

## Checklist:

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -24,6 +24,9 @@ jobs:
         with:
           node-version: 20
 
+      - name: Pin npm version 📌
+        run: npm install -g npm@11.17.0
+
       - name: Install 🔧
         run: npm ci
 
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -54,6 +54,9 @@ jobs:
         node-version: 20
         registry-url: https://registry.npmjs.org/
 
+    - name: Pin npm version 📌
+      run: npm install -g npm@11.17.0
+
     - name: Install locked dependencies 🔧
       run: npm ci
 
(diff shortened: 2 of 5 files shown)


dbernstein (approved):
🚀 ☘️ 🚀

PR #301, opened 2026-06-26
Revert dependabot's bootstrap v5 update back to v3 (PP-4661)

tdilauro:
## Description

- Reverts dependabot's `bootstrap` upgrade (from v3 -> v5) on #298.
- Adds dependabot configuration to pin `bootstrap` to versions less than v4.

## Motivation and Context

Bootstrap v5 is a radical departure from v3, dropping support for many components. This app depends on components and configuration in v3.

The dependabot configuration update is needed to prevent dependabot doing the same thing again.

[Jira PP-4661]

## How Has This Been Tested?

- Manual testing using `npm run dev-server` against Minotaur.
- All tests pass locally and in CI, but they also did with the wrong `bootstrap` version.

## Checklist:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,11 @@ updates:
     versioning-strategy: increase-if-necessary
     cooldown:
       default-days: 4
+    ignore:
+      # We currently support only v3 of Bootstrap. Prevent PRs for v4.0.0 or higher.
+      # See https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/configure-version-updates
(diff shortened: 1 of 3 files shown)


jonathangreen (approved):
Looks good! I think the ignore can also be configured by commenting on the PR in github.

tdilauro:
With the dependabot comments, it looked like I could ignore the dependency or the major version. I didn't want to ignore the dependency entirely, but just non-v3 versions, in case there is a v3 fix (though now I see that Claude believes this will not come). I wanted to avoid v4, as well, so the major version was too specific.

PR #302, opened 2026-06-27
Use Node v24 (PP-4612)

tdilauro:
## Description

Updates the project to require Node.js 24 as the minimum (and only tested) version. The CI test matrix, publish and docs workflows, local version manager files (`.nvmrc`, `.node-version`), `package.json` engines constraint, and developer documentation are all updated to reflect this.

## Motivation and Context

The project has been verified to work correctly on Node.js 20, 22, and 24. Now that Node 24 is the current LTS, we are dropping support for older versions and pinning the project to Node 24. The upper bound remains `<25.0.0` since Node 25 has not yet been tested. This gets us to a supported LTS, eliminates the need to support older behaviors and for some polyfills, and allows us to work toward the support for the newly-minted v26 LTS.

Note: after merging, the branch protection required status checks should be updated — remove **Test (20)** and **Test (22)**. **Test (24)** is already present and can remain as-is.

[Jira PP-4612]

## How Has This Been Tested?

The CI test matrix previously included Node 20, 22, and 24 and all passed. This change narrows for now the matrix to Node 24 only, which was already green.

## Checklist:


jonathangreen (approved):
🤘 this is great!

PR #310, opened 2026-07-09
Rewrite the enzyme-free legacy tests as jest tests

jonathangreen:
## Description

Rewrites the twelve legacy test suites that never imported enzyme — the nine reducers, the `Admin` model, `editorAdapter`, and `sharedFunctions` — as jest tests. They move from `src/*/__tests__/` into `tests/jest/` mirroring `src/`, which is where new tests live (`sharedFunctions` lands under `tests/jest/utils/`, matching the module it exercises).

Because these suites only used chai (never enzyme), the rewrite was mechanical — 424 assertions translated by pattern:

```
.to.equal(x)        -> .toBe(x)
.to.deep.equal(x)   -> .toStrictEqual(x)
.to.be.ok           -> .toBeTruthy()
.not.to.be.ok       -> .toBeFalsy()
.to.have.property() -> .toHaveProperty()
.to.be.true/false   -> .toBe(true/false)
.to.be.undefined    -> .toBeUndefined()
```

`deep.equal` maps to `toStrictEqual` rather than `toEqual`: chai's `deep-eql` fails when one side has an own key set to `undefined`, and `toEqual` does not. That changed one assertion in `editorAdapter`, where the adapter deliberately flattens `Contributor` instances into plain objects and the test compared against the instances — chai ignores prototypes, `toStrictEqual` does not. It now asserts the shape the adapter actually produces, which is what was meant. Six `expect(x).to.have.property(k).that.deep.equals(v)` chains in `customListEditor` collapse to `expect(x).toHaveProperty(k, v)`.

101 legacy suites remain, all of them enzyme.

## Motivation and Context

Part of the ongoing mocha/chai/enzyme → jest/RTL migration. Clearing the enzyme-free suites first shrinks the legacy project without touching source, and each rewrite is a straight assertion translation with no behavioral risk. The enzyme suites — which need real component-rendering replacements — are left for follow-up; retiring the last of them is what unblocks React 17+.

## How Has This Been Tested?

`npm test` — full suite green. Test count is unchanged at 165 for these files and 1431 overall. Coverage is identical (6596/6999 lines, 94.24%) with no previously-covered line lost; none of the source files changed, so the gate compared them line by line.

## Checklist:


tdilauro (approved):
Looks good!

I really just gave this one a quick scan because of the "mechanical" nature of the changes.

PR #311, opened 2026-07-09
Rewrite actions-test as a jest test

jonathangreen:
## Description

Rewrites `actions-test` — the last legacy suite that never imported enzyme — as a jest test, moving it from `src/__tests__/actions-test.ts` to `tests/jest/actions.test.ts`. 353 assertions convert mechanically: chai to jest's `expect`, and sinon's bare `stub()` spy-props to `jest.fn()`:

```
.callCount     -> .mock.calls.length
.args[i][j]    -> .mock.calls[i][j]
```

There were no sandboxes, fake timers, or `stub(obj, "method")` forms to unpick.

It also moves from `require("fetch-mock")` to `import * as fetchMock from "fetch-mock-jest"`, matching the seven jest tests that already use it. The bare `require` was load-bearing: fetch-mock's runtime export is the mock itself, but its typings declare a *default* export, so with `esModuleInterop` off only an untyped `require` compiles. `fetch-mock-jest` declares `export =`, so the import type-checks — which immediately caught that `fetchArgs[0][1].body` is typed `BodyInit` and needs a cast before `.get("force")`.

That leaves `fetch-mock` unimported anywhere at the top level, so it's dropped from `devDependencies`. `fetch-mock-jest` carries its own nested v9, so the unused v10 was only shadowing it.

100 legacy suites remain, all of them enzyme.

## Motivation and Context

Part of the ongoing mocha/chai/enzyme → jest/RTL migration. With this suite converted, every legacy test that didn't depend on enzyme is now a jest test; what remains is purely the enzyme-rendering suites, whose retirement is what unblocks React 17+.

## How Has This Been Tested?

`npm test` — full suite green. 32 tests in this suite before and after; 1431 overall, 6596/6999 lines, 94.24%, no previously-covered line lost.

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -69,7 +69,6 @@
         "eslint-plugin-jsx-a11y": "^6.10.2",
         "eslint-plugin-react": "^7.37.5",
         "eslint-plugin-react-hooks": "^7.1.1",
-        "fetch-mock": "^10.0.7",
         "fetch-mock-jest": "^1.5.1",
         "fetch-ponyfill": "^7.1.0",
         "file-loader": "^6.2.0",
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
Looks good! 🚀 

Again, mostly a quick scan, since mostly "mechanical" changes.

PR #312, opened 2026-07-09
Rewrite SaveButton and UpdatingLoader tests with React Testing Library

jonathangreen:
## Description

Rewrites the `SaveButton` and `UpdatingLoader` tests from enzyme to React Testing Library — the first two enzyme suites to go, chosen to establish the pattern on the simplest cases: one class component with a click handler, one pure function component. They move from `src/components/__tests__/` to `tests/jest/components/`.

The translation pattern:

```
mount + wrapper.text()        -> render + getByRole / getByText
wrapper.setProps({ ... })     -> a fresh render with the new props
wrapper.simulate("click")     -> await userEvent.click(...)
sinon stub() + .callCount     -> jest.fn() + toHaveBeenCalledTimes
wrapper.find(".some-class")   -> container.querySelector(".some-class")
```

Querying by role rather than by class is the point of the exercise, but two assertions have no accessible handle: the `.updating-loader` wrapper div and the Font Awesome `<i>` spinner. Those keep a class selector, with a comment saying why.

The disabled-button test gets stronger on the way across. Enzyme's `simulate` dispatches straight to React's handler, so it only passed because React itself declines to fire `onClick` on a disabled button; `userEvent.click` models what a browser actually does.

98 legacy suites remain.

## Motivation and Context

Part of the ongoing mocha/chai/enzyme → jest/RTL migration. With the enzyme-free suites already converted (#310, #311), this begins the harder half: the enzyme component suites, whose retirement is what unblocks React 17+. Starting with the two simplest components establishes the RTL translation pattern for the ones that follow.

## How Has This Been Tested?

`npm test` — full suite green. Per-file line coverage is unchanged (SaveButton 8/8, UpdatingLoader 3/3), as is the total (6596/6999). 8 tests before, 8 after.

## Checklist:


tdilauro (approved):
👍🏽

PR #313, opened 2026-07-09
Rewrite the shallow presentational tests with React Testing Library

jonathangreen:
## Description

Rewrites the 15 suites that shallow-rendered and asserted "child X received prop Y" — the page shells (`AccountPage`, `ConfigPage`, `CustomListPage`, `DashboardPage`, `LanePage`, `SetupPage`, `WelcomePage`), the containers (`BookDetailsContainer`, `EntryPointsContainer`, `ManagePatrons`), and the small presentational pieces (`BookDetails`, `EntryPointsTabs`, `PatronInfo`, `SingleStatListItem`, `TroubleshootingTabContainer`) — from enzyme to React Testing Library, moving them into `tests/jest/components/`.

RTL has no shallow renderer, and asserting a child's props is exactly what it exists to discourage, so these render for real and assert the DOM the child produces. Where a child fetches on mount or needs react-router context that `renderWithProviders` doesn't supply, it's mocked to a marker — `Header`/`Footer` (chrome, need a router), react-router `Link`, and the connected tab containers (`BookDetailsTabContainer`, `ConfigTabContainer`, `CustomLists`, `Stats`) that mount network-fetching panels. Everything else renders for real.

Some page-shell assertions necessarily check that a marker child received non-serializable props (a store, a callback) via presence data-attributes; the original identity (`===`) checks had no DOM-observable effect and collapse to presence. That's the floor for a pure forwarding container whose real child can't mount without the network.

83 legacy suites remain, all enzyme.

## Motivation and Context

Part of the ongoing mocha/chai/enzyme → jest/RTL migration. This is the largest tranche so far — the shallow-render-and-check-props suites — and clearing it leaves only the deeper enzyme suites. Retiring the last of them is what unblocks React 17+. `userEvent` usage follows the `userEvent.setup()` convention established across `tests/jest/`.

## How Has This Been Tested?

`npm test` — full suite green. Coverage rises rather than holds — 94.24% → 94.48%, +24 lines — because rendering the real children exercises code shallow rendering never reached (`SetupPage`'s `IndividualAdmins` and `LanePage`'s `Lanes` even pull in action creators). No previously-covered line lost.

## Checklist:


tests/jest/components/EntryPointsContainer.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+jest.mock("react-router", () => ({
+  ...jest.requireActual("react-router"),
+  Link: (props) => (
+    <div data-testid="Link" data-to={props.to}>
+      {props.children}
+    </div>
+  ),
+}));
+
+import EntryPointsContainer from "../../../src/components/EntryPointsContainer";
```
tdilauro, line 29: Minor - Just wondering why this import is not up at the top.

No need to change this now, as I know there are a lot of PRs stacked here and don't want to create any merge conflicts for something like this that can be cleaned up later, if needed.

tests/jest/components/DebugAuthentication.test.tsx
tdilauro, line 1: Just linting?

tests/jest/components/EditableConfigList.test.tsx
tdilauro, line 1: Just linting?

tdilauro (approved):
This looks good and is approved, but there were a bunch of files with just linting changes. I flagged a couple, but there were a bunch more.

tests/jest/components/DebugAuthentication.test.tsx
jonathangreen: Yes. I'm not sure why the linter decided to pick up.

tests/jest/components/EntryPointsContainer.test.tsx
jonathangreen: Claude seems to prefer the style of:
- General imports at the top of the file
- Any mocks
- Import whats being tested
- Do the testing

That pattern is used a number of times in these PRs where there are mocks. Its not necessary, I thought maybe it was needed for the mocks to be active at import time, but I changed the order and it doesn't seem like it is. So its just a style choice. 

I think I'll leave it as is for now, but we can re-order the imports after if we want to.

tests/jest/components/DebugAuthentication.test.tsx
jonathangreen: I figured out what was up here. Prettier isn't actually run in CI, its just run by husky in a pre-commit hook. So the prettier version has been bumped, and there are about 40 files that need formatting changes, but its only happening when the files get touched. 

I'm going to leave these for now, but I'll work on a follow up after to make the linter situation better, so this stops happening.

PR #314, opened 2026-07-09
Rewrite the mount-based presentational tests with React Testing Library

jonathangreen:
## Description

Rewrites 24 of the 28 mount-based enzyme suites to React Testing Library: the AdvancedSearch filter widgets (rendered inside a real `DndProvider` — HTML5Backend works under jsdom), the small leaf components and buttons, the self-test and manage-patrons panels, the search/sidebar/services views, three of the tab containers, `Footer`, and the app bootstrap (`index.tsx`).

- The tab containers navigate via `router.push` rather than internal state, so "active tab" is asserted through the visible panel; where a tab panel fetches on mount and is incidental to a navigation test, it's mocked to a marker.
- `LibraryStats` is deleted outright rather than ported — `tests/jest/components/Stats.test.tsx` already covers `LibraryStats.tsx` at 100% of lines and tests more (the query hook, caching, sysadmin gating).
- The bootstrap test keeps its structural shape: it spies on `ReactDOM.render`, constructs `CirculationAdmin` in each mode, and asserts the captured element tree contains `SetupPage` / a `Router` — the analog of the original enzyme `find()`, since what those roots render is covered by their own tests.

`userEvent` usage follows the `const user = userEvent.setup()` convention used across `tests/jest/`.

**Deferred, with their enzyme originals retained:** `ConfigTabContainer` and `BookDetailsTabContainer` (their full mounts are the sole cover for eleven connected panels' `mapStateToProps`/`mapDispatchToProps` wiring, restored properly when those panels migrate to render their connected exports) and `AdvancedSearchBuilder` (its one uncovered line is a react-dnd move callback reachable only through a real drop). Mocking those panels now would have dropped 77 lines; the coverage gate caught it.

## Motivation and Context

Part of the mocha/chai/enzyme → jest/RTL migration — the components enzyme actually mounted, the largest tranche. Retiring the last enzyme suites is what unblocks React 17+.

## How Has This Been Tested?

`npm test` — full suite green. Coverage rises to 94.48% (+24 lines over the baseline) with no previously-covered line lost; rendering real leaf children reaches branches the shallow and mount-with-stubbed-children tests did not.

## Checklist:


tests/jest/components/AdvancedSearchBooleanFilter.test.tsx
```diff
@@ -0,0 +1,233 @@
...
+    expect(options[0]).toHaveTextContent(
+      "All of these filters must be matched:"
+    );
+    expect((options[0] as HTMLOptionElement).selected).toBe(true);
+
+    expect(options[1]).toHaveValue("or");
+    expect(options[1]).toHaveTextContent(
+      "Any of these filters may be matched:"
+    );
+    expect((options[1] as HTMLOptionElement).selected).toBe(false);
```
tdilauro: Minor - A couple of these (the `true`/`false`) assertions are unneeded because of the `expect(select).toHaveValue("and");` at the end of the test:


```suggestion

    expect(options[1]).toHaveValue("or");
    expect(options[1]).toHaveTextContent(
      "Any of these filters may be matched:"
    );
```

tests/jest/components/CirculationEventsDownloadForm.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    // form's Download button and the footer's Close button are the reusable
+    // `Button`s (class "btn"), matching the two the legacy test counted.
+    const buttons = document.querySelectorAll<HTMLButtonElement>("button.btn");
+    expect(buttons).toHaveLength(2);
+    expect(buttons[0]).toHaveTextContent("Download");
+    expect(buttons[1]).toHaveTextContent("Close");
+
+    expect(
+      screen.getByRole("button", { name: "Download" })
+    ).toBeInTheDocument();
```
tdilauro, line 29: Minor - This is duplicating the check for one of the buttons above, but it is the way I would prefer (over the approach above) to check for the buttons with Jest/RTL.

```
expect(screen.getByRole("button", { name: "Download" })).toBeInTheDocument();
expect(screen.getByRole("button", { name: "Close" })).toBeInTheDocument();
```

tests/jest/components/CustomListsSidebar.test.tsx
```diff
@@ -0,0 +1,246 @@
...
+import CustomListsSidebar from "../../../src/components/CustomListsSidebar";
+
+// Render react-router's Link as a marker anchor so we can assert its target
+// (`data-to`), text, and className without needing a Router in the test.
+jest.mock("react-router", () => ({
+  ...jest.requireActual("react-router"),
+  Link: (props) => (
+    <div data-testid="Link" data-to={props.to} className={props.className}>
+      {props.children}
+    </div>
```
tdilauro: Minor - It would make more sense for this to be an `<a>` than a `<div>`.


```suggestion
    <a data-testid="Link" data-to={props.to} className={props.className}>
      {props.children}
    </a>
```

tdilauro (approved):
Looks good!

A few minor things, but no showstoppers.

tests/jest/components/AdvancedSearchBooleanFilter.test.tsx
jonathangreen: Done.

tests/jest/components/CustomListsSidebar.test.tsx
jonathangreen: I switched this to an `<a>`, but I also needed to add a `href`. Otherwise linting fails on `jsx-a11y/anchor-is-valid`.

tests/jest/components/CirculationEventsDownloadForm.test.tsx
jonathangreen: `expect(screen.getByRole("button", { name: "Close" })).toBeInTheDocument()` doesn't work here, because the headers `x` dismiss button is also called `Close`. So using that causes `getByRole` to throw because it matched two elements. 

I made some changes here though to clean up the test a little bit and scope the `getByRole` down to the footer.

PR #318, opened 2026-07-13
Rewrite custom-list test suites with React Testing Library

jonathangreen:
## Description

Rewrites the custom-list and list-entry editing test suites with React Testing Library.

Migrated suites (old enzyme test in `src/components/__tests__/` removed, RTL test added/expanded in `tests/jest/components/`):

- CustomListEntriesEditor
- CustomLists
- LaneCustomListsEditor
- PairedMenus
- InputList

Migration notes (carried over from #316, for the suites in this PR):

- `CustomListEntriesEditor` was the largest suite in the whole migration (originally 1461 LOC).
- `CustomListEntriesEditor` and `LaneCustomListsEditor` use `react-beautiful-dnd`, which can't drag in jsdom. `DragDropContext` is mocked to capture `onDragStart`/`onDragEnd`, which the tests then invoke with synthetic drag results — both suites reach 100% coverage this way.
- Enzyme live-instance-method spies become observable behavior — asserting which entries/rows are removed vs. retained — instead of assertions on class internals.
- Where a suite covers a connected panel, it renders the connected default export so `mapStateToProps`/`mapDispatchToProps` stay covered.

This is one of four PRs splitting the oversized #316 into independently reviewable pieces, grouped by feature area. Each split branches from `main` and touches a disjoint set of component test files, so the four can be reviewed and merged in any order.

## Motivation and Context

#316 rewrote the HARD-tier and remaining merge test suites from enzyme to React Testing Library in a single ~11k-line change, which is impractical to review. Splitting it by feature area keeps each PR at a reviewable size while preserving the same net result. These were among the hardest suites in the mocha/chai/enzyme → jest/RTL migration — enzyme mounted them and poked class internals, so the rewrites re-express those assertions as DOM and behavior checks.

## How Has This Been Tested?

`npx jest --selectProjects unit` over the migrated suites on this branch: **5 suites / 109 tests pass**. No application source changes — test files only.

## Checklist:


tests/jest/components/CustomListEntriesEditor.test.tsx
```diff
@@ -0,0 +1,1241 @@
...
+        entryCount: 2,
+        deleteAllEntries,
+      })
+    );
+
+    expect(entriesHeadingText(container)).toBe(
+      "List Entries: Displaying 1 - 2 of 2 books"
+    );
+
+    await user.click(screen.getByRole("button", { name: "Delete" }));
```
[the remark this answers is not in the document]
jonathangreen: The SVG does not contribute title text to the button name. So this is a false alarm. Resolving.

tests/jest/components/LaneCustomListsEditor.test.tsx
```diff
@@ -0,0 +1,474 @@
...
+  });
+
+  it("removes a list when dragged from current to available", () => {
+    const onUpdate = jest.fn();
+    // Fresh customListIds: the component's add() pushes onto the array in place,
+    // so the shared `props.customListIds` can be mutated by the test above.
+    render(
+      <LaneCustomListsEditor
+        {...props}
+        customListIds={[2]}
```
tdilauro: The first test here is mutating the `customListIds` in `props`. Most of the tests use the original value, but all have to reset it. Might be better to make props with a factory function rather than setting it as a `const` (which only protects the top-level of the object).

tdilauro (approved):
Looks good!

One suggestion, but not a showstopper and could even be addressed later.

tests/jest/components/LaneCustomListsEditor.test.tsx
jonathangreen: Good call

PR #321, opened 2026-07-13
Rewrite admin, announcement and shared-primitive test suites with React Testing Library

jonathangreen:
## Description

Rewrites the admin, discovery, announcement, and shared form/presentational primitive test suites with React Testing Library.

Migrated suites (old enzyme test in `src/components/__tests__/` removed, RTL test added/expanded in `tests/jest/components/`):

- IndividualAdminEditForm
- IndividualAdmins
- ChangePasswordForm
- Header
- LibraryRegistration
- DiscoveryServices
- NeighborhoodAnalyticsForm
- AnnouncementForm
- AnnouncementsSection
- TextWithEditMode
- EditableInput
- ToolTip

Migration notes (carried over from #316, for the suites in this PR):

- Prototype stubs (`EditableInput.prototype.getValue`, `TextWithEditMode`) become real `userEvent.type` that reads back through the component's ref API; `state()`/`instance()` reads become DOM assertions.
- `EditableInput` and `IndividualAdmins` merge into existing jest suites rather than adding new files.
- Where a suite covers a connected panel (e.g. `DiscoveryServices`, `Header`), it renders the connected default export so `mapStateToProps`/`mapDispatchToProps` stay covered.
- `userEvent` usage follows the `const user = userEvent.setup()` convention used across `tests/jest/`.

This is one of four PRs splitting the oversized #316 into independently reviewable pieces, grouped by feature area. Each split branches from `main` and touches a disjoint set of component test files, so the four can be reviewed and merged in any order.

## Motivation and Context

#316 rewrote the HARD-tier and remaining merge test suites from enzyme to React Testing Library in a single ~11k-line change, which is impractical to review. Splitting it by feature area keeps each PR at a reviewable size while preserving the same net result. These were among the hardest suites in the mocha/chai/enzyme → jest/RTL migration — enzyme mounted them and poked class internals, so the rewrites re-express those assertions as DOM and behavior checks.

## How Has This Been Tested?

`npx jest --selectProjects unit` over the migrated suites on this branch: **12 suites / 153 tests pass**. No application source changes — test files only.

## Checklist:


tests/jest/components/IndividualAdmins.test.tsx
```diff
@@ -388,3 +385,32 @@ describe("IndividualAdmins - role association disclosure", () => {
     expect(container.querySelectorAll(".associated-items")).toHaveLength(0);
   });
 });
+
+// Exercises the connected default export so mapStateToProps/mapDispatchToProps
+// are covered. The RBAC gating is already covered by the disclosure tests above
+// and by tests/jest/businessRules/roleBasedAccess.test.ts.
```
tdilauro: This comment is not correct. The tests in `tests/jest/businessRules/roleBasedAccess.test.ts` cover inventory reports, the QuickSight link, and the statistics bar chart; and the disclosure tests above are only covering role _display_, not actual admin user configuration.

Change requested - It seems that a bunch of the access control tests from the old `src/components/__tests__/IndividualAdmins-test.tsx` were dropped during the migration. These should be added back in.

tdilauro (commented):
This mostly looks good, but it appears that a tranche of role-based tests for `IndividualAdmins` were dropped during the migration.

tests/jest/components/IndividualAdmins.test.tsx
jonathangreen: Thanks for catching this Tim. I went though and cleaned this up the comment and the tests. 

I also sent Claude at this PR again with a prompt to re-review it tests that got dropped in the migration, and it came up with a few more gaps. The new commit I pushed covers both your feedback and these additional test cases.

tdilauro (approved):
Looks good! 🏁 🍾

PR #320, opened 2026-07-13
Rewrite service and collection configuration test suites with React Testing Library

jonathangreen:
## Description

Rewrites the service and collection configuration test suites with React Testing Library.

Migrated suites (old enzyme test in `src/components/__tests__/` removed, RTL test added/expanded in `tests/jest/components/`):

- ServiceEditForm
- EditableConfigList
- SelfTests
- ProtocolFormField
- Collections

Migration notes (carried over from #316, for the suites in this PR):

- `ServiceEditForm` was one of the largest suites in the migration (originally 1087 LOC).
- Enzyme live-instance-method spies become observable behavior — e.g. asserting the submitted `FormData` (via the shared FormData shim) rather than that a save method was called.
- Tiny test subclasses drive branches that the base `EditableConfigList` method short-circuits (`Collections` extends `EditableConfigList`).
- Where a suite covers a connected panel, it renders the connected default export so `mapStateToProps`/`mapDispatchToProps` stay covered.

This is one of four PRs splitting the oversized #316 into independently reviewable pieces, grouped by feature area. Each split branches from `main` and touches a disjoint set of component test files, so the four can be reviewed and merged in any order.

## Motivation and Context

#316 rewrote the HARD-tier and remaining merge test suites from enzyme to React Testing Library in a single ~11k-line change, which is impractical to review. Splitting it by feature area keeps each PR at a reviewable size while preserving the same net result. These were among the hardest suites in the mocha/chai/enzyme → jest/RTL migration — enzyme mounted them and poked class internals, so the rewrites re-express those assertions as DOM and behavior checks.

## How Has This Been Tested?

`npx jest --selectProjects unit` over the migrated suites on this branch: **5 suites / 134 tests pass**. No application source changes — test files only.

## Checklist:

- [x] All new and existing tests passed.

PR #322, opened 2026-07-13
Adopt a color-picker value that arrives after mount

jonathangreen:
## Description

`ColorPicker` seeds its state from `props.value` in the constructor and never updates it again — it has no props→state sync, and `ProtocolFormField.renderColorPickerSetting` passes no `key`, so the instance is never remounted either. A `value` that arrives from the parent *after* mount is therefore ignored, and both the hidden input and `getValue()` keep returning the value the picker was first constructed with.

This adds a guarded props→state sync in `componentDidUpdate`, matching the pattern `InputList` already uses (`InputList.tsx:72`).

The guard (`prevProps.value !== this.props.value`) is load-bearing in two ways: it stops an unrelated re-render from discarding a color the admin has just picked, and — since this is `componentDidUpdate` — an unguarded `setState` would recurse (`Maximum update depth exceeded`).

## Motivation and Context

This is a real bug, not just missing coverage.

A configuration form typically mounts before its setting data has loaded, so the color field initializes from `setting.default`. When the saved setting then arrives from the parent, `ColorPicker` ignores it and keeps the default. Because `getValue()` reads that same stale state, **submitting the form silently overwrites the admin's saved color with the default**, with no indication in the UI.

Found while reviewing the `ProtocolFormField` test migration in #320 — a reviewer asked for a hidden color-picker rerender case, and writing it surfaced the underlying defect. The fix is kept out of #320 because that PR is deliberately test-only.

## How Has This Been Tested?

Two regression tests in `ProtocolFormField.test.tsx`, each verified to fail without the behavior it covers:

- **`adopts a color-picker value that arrives after mount`** — mounts with no value (so the picker starts from the default), then rerenders with the saved value and asserts `getValue()` returns it. Fails against the unfixed component (`Expected: "#123456"`, `Received: "#aaaaaa"`) — this is the bug above.
- **`keeps a color the admin picked when the parent re-renders`** — clicks a swatch, then rerenders with an unchanged `value` prop and asserts the pick survives. Fails if the sync is made unconditional, so it pins the guard.

Full suite green: **142 suites / 1428 tests** across both jest projects, including the existing enzyme `ColorPicker-test`.

## Checklist:

### DIFF
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -21,6 +21,17 @@ export default class ColorPicker extends React.Component<
     this.handleChange = this.handleChange.bind(this);
   }
 
+  // The form can mount before its setting data has loaded, in which case this
+  // picker initializes from the setting's default. Adopt the real value when the
+  // parent supplies it, or the field would keep submitting that stale default.
+  // Only adopt it when the incoming value actually changes, so that an unrelated
+  // re-render does not discard a color the admin has just picked.
+  componentDidUpdate(prevProps: ColorPickerProps) {
+    if (prevProps.value !== this.props.value) {
+      this.setState({ value: this.props.value });
+    }
+  }
+
   render(): JSX.Element {
     return (
       <div className="color-picker">
--- a/tests/jest/components/ProtocolFormField.test.tsx
+++ b/tests/jest/components/ProtocolFormField.test.tsx
@@ -457,6 +457,100 @@ describe("ProtocolFormField — setting types", () => {
     );
   });
 
+  it("adopts a color-picker value that arrives after mount", () => {
+    const ref = React.createRef<ProtocolFormField>();
+    const setting = {
+      ...baseSetting,
+      hidden: true,
+      type: "color-picker",
+      default: "#aaaaaa",
+    };
+    const { rerender } = render(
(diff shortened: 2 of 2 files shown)


jonathangreen:
Both bots flagged the same race from different angles, Claude suggested gating the sync on `this.state.value === prevProps.value`, and Greptile suggested having handleChange record that the field was edited. I rejected both of these because we cannot hit this race in our code. `EditableConfigList.itemToEdit()` returns `null` until the fetched data contains the item, and the `EditForm` only renders when it's truthy, so in both usages of this field (`ServiceEditForm` and `LibraryEditForm`) the picker mounts with the saved value already in props, never showing `setting.default` while a value is in flight.

`ColorPicker` cannot distinguish a late-arriving value for the _current_ item (where you'd want to keep the pick) from a switch to a _different_ item (where you must adopt the new value). Both reach the component as nothing more than "props.value changed X→Y". I added a test for the item-switch case and ran it against both suggestions as written, which caused the test to fail. So the contract I'm keeping, now pinned by tests, is:

- an unchanged value prop never disturbs the field (this is the keeps a color the admin picked when the parent re-renders case), and
- a changed value prop from the parent always wins (the item-switch case).

That's the same semantics `EditableInput` already has, so the two fields behave consistently.

PR #325, opened 2026-07-16
Rewrite the remaining container tests with React Testing Library

jonathangreen:
## Description

Replaces the last four enzyme container suites — `AdvancedSearchBuilder`, `BookDetailsTabContainer`, `ConfigTabContainer`, `SelfTestsTabContainer` — with React Testing Library, and rounds out RTL coverage for `Libraries` (new), `CatalogServices`, `CustomListsForBook` and `MetadataServices`.

Each of these containers renders connected panels that fetch on mount, so the panels are mocked to markers echoing the props the legacy tests asserted. That keeps the suites about the containers' own behavior — the tab set, tab switching, and prop threading — and it reaches callbacks that jsdom otherwise can't drive, notably `AdvancedSearchBuilder`'s `onMove`, which is only reachable through a real react-dnd drag.

Every case from the deleted suites is carried over, and two are now stricter than the originals:

- **The router assertion actually checks the tab key.** The legacy test compared `push` against `mockRouterContext`'s `pathFor`, which takes only two parameters and ignores the third. The old assertion therefore passed no matter which tab key the container sent. The new test supplies a three-argument `pathFor` and asserts the real `dataset.tabkey`.
- **`getNames()` is tested through observable output.** The legacy suite reached into `wrapper.instance().getNames(...)`. That is replaced by asserting the store key, type and link name each category actually receives — the same derivation, without reaching for the instance.

## Motivation and Context

Continues the enzyme-to-RTL migration. Enzyme has no adapter beyond React 16, so retiring these suites is on the path to unblocking React 17+; per `CLAUDE.md`, the legacy jest project and its jsdom environment can be deleted along with the last enzyme test.

## How Has This Been Tested?

`npx jest --selectProjects=unit` over the eight touched suites: 49 tests passing. `npm run lint:js` and prettier are clean.

The two assertions added during review were verified by mutation rather than by passing alone — each was confirmed to fail against the regression it targets:

- Pointing every `SelfTestsCategory` at a hardcoded store key (a type-compatible bug the compiler does not catch) fails the store-key test.
- Making `mapStateToProps` stop threading `bookData` fails the connected-mount test.

Both source files were restored afterward; this PR touches no `src/` file other than deleting the four legacy suites.

## Checklist:


tests/jest/components/Libraries.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+                uuid: "uuid-nypl",
+                short_name: "nypl",
+                name: "New York Public Library",
+                settings: {},
+              },
+            ],
+          }),
+          { status: 200, headers: { "Content-Type": "application/json" } }
+        )
+    );
```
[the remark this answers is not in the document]
jonathangreen: Declining this one. There are no MSW handlers for the fetch spy to bypass. `renderWithProviders` (`tests/jest/testUtils/withProviders.tsx`) contains no MSW code at all, and the `unit` project's `setupFilesAfterEnv` is `tests/jest/jest-setup.ts`, whose entire contents are `import "@testing-library/jest-dom";`.

jonathangreen:
Re Claudes review: 
> This test is named/commented as covering the bookUrl-change re-sync, but its assertion is satisfied by prop threading alone and does not exercise the lifecycle it targets.

- Nothing reads `custom_lists`. `state.customLists` flows only into `setting.custom_lists`. Neither `ProtocolFormField` nor `InputList` mentions it.
- The rendered column comes from props. `value` is `this.props.customListsForBook.map(l => l.name)`, which is why "list 1" appears on rerender regardless of the lifecycle.

So `state.customLists` is write-only: the constructor seeds it, the lifecycle re-syncs it, and it reaches a setting field no one consumes. I think the solution here is to delete the dead code, but I don't want to bring that into this PR, so I think I'm going to leave this as is for now, and make a follow up the cleans up the dead code.

PR here: https://github.com/ThePalaceProject/circulation-admin/pull/327

tests/jest/components/CatalogServices.test.tsx
```diff
@@ -61,3 +61,40 @@ describe("CatalogServices", () => {
     );
   });
 });
+
+import CatalogServicesConnected from "../../../src/components/CatalogServices";
+import { renderWithProviders } from "../testUtils/withProviders";
+import buildStore from "../../../src/store";
```
tdilauro: Any reason not to have these imports at the top of the file?

tests/jest/components/CustomListsForBook.test.tsx
```diff
@@ -312,4 +312,37 @@ describe("CustomListsForBook", () => {
     expect(screen.getByRole("button", { name: "Delete" })).toBeDisabled();
     expect(screen.getByRole("button", { name: "Add" })).toBeDisabled();
   });
+
+  it("resyncs its lists when the book url changes", () => {
```
tdilauro, line 316: There are a couple of issues with the test, so I pulled down the branch and did some testing:
- Both "list 1" and "list 2" are on the page as options, so `getByText` finds them and both checks pass. There's not a negative assertion for the other list on each render, so the test passes.
  - Testing with the link role and making a negative assertion gets us closer. I'll add suggestions for these below.
- The test name is misleading. It's not the `bookUrl` (or even the `book`) changing. When I left them the same the test continued to pass.
  - It's really the change to `customListsForBook` that results in the change across renders.
  - We should leave the book and book URL unchanged to avoid confusion about what is under test.

tests/jest/components/CustomListsForBook.test.tsx
```diff
@@ -312,4 +312,37 @@ describe("CustomListsForBook", () => {
...
+    const { rerender } = render(
+      <UnconnectedCustomListsForBook
+        {...commonProps}
+        book={{ id: "a", title: "Book A" } as any}
+        bookUrl="works/book-a"
+        customListsForBook={listsForBook as any}
+      />
+    );
+    // list 2 (the book's only list) is in the current column.
+    expect(screen.getByText("list 2")).toBeInTheDocument();
```
tdilauro, line 335:
```suggestion
    expect(screen.getByRole("link", { name: "list 2" })).toBeInTheDocument();
    expect(
      screen.queryByRole("link", { name: "list 1" })
    ).not.toBeInTheDocument();
```

tests/jest/components/CustomListsForBook.test.tsx
```diff
@@ -312,4 +312,37 @@ describe("CustomListsForBook", () => {
...
+    // A new book url re-syncs the current lists from the new customListsForBook.
+    rerender(
+      <UnconnectedCustomListsForBook
+        {...commonProps}
+        book={{ id: "b", title: "Book B" } as any}
+        bookUrl="works/book-b"
+        customListsForBook={[allCustomLists[0]] as any}
+      />
+    );
+    expect(screen.getByText("list 1")).toBeInTheDocument();
```
tdilauro, line 346:
```suggestion
    expect(screen.getByRole("link", { name: "list 1" })).toBeInTheDocument();
    expect(
      screen.queryByRole("link", { name: "list 2" })
    ).not.toBeInTheDocument();
```

tdilauro (approved):
This looks good!

A few comments below:
- a test that is not doing what it purports to do and is actually broken; and
- some imports in the middle of a file.

I'm going to go ahead and approve, since neither of those things will cause problems and it looks like most of the malfunctioning test is covered elsewhere.

tests/jest/components/CustomListsForBook.test.tsx
tdilauro: Nevermind. This is fixed on #327.

tests/jest/components/CatalogServices.test.tsx
jonathangreen: good call, moved.

PR #326, opened 2026-07-16
Remove enzyme, chai, and sinon; document the test setup

jonathangreen:
## Description

Stacked on #325 — please review and merge that first; this PR targets `chore/rewrite-container-tests-rtl` and will retarget to `main` once #325 lands.

With the last enzyme suites rewritten in #325, nothing depends on the old test infrastructure any more. This tears it out and brings the docs in line with what's left.

**Teardown**

- Drop the `enzyme`, `enzyme-adapter-react-16`, `chai`, and `sinon` devDependencies.
- Collapse `jest.config.js` from two projects (`unit` + `legacy`) to a single config, and delete the machinery the `legacy` project needed: `tests/jest/legacyJsdomEnvironment.js` and `tests/jest/legacy-setup.ts`.
- Delete `src/components/__tests__/routing.ts`, a helper with no remaining callers.

**Move the last test code out of `src`**

The two shared fixtures (`genreData`, `classificationsData`) were never tests — just data the old suites shared — and four RTL suites still import them. They move to `tests/jest/fixtures/`, which leaves `src` free of test code entirely. With `src/**/__tests__/**` no longer matching anything, that exclusion is dropped from both `collectCoverageFrom` in `jest.config.js` and the `ignore` list in `.github/codecov.yml`.

**Docs**

Rewritten to describe the setup we have: where tests live, the RTL helpers in `testUtils`, how the network gets stubbed, and the gotchas that cost real time (render the connected export or lose `connect` coverage; settle on-mount fetches or the run fails). `tests/jest/README.md` is deleted rather than rewritten — everything in it was already in README.md or CLAUDE.md. 

## Motivation and Context

The enzyme-to-RTL migration is complete as of #325. Enzyme was the thing pinning us to React 16 — `enzyme-adapter-react-16` has no successor for React 17+ — so removing it is what unblocks a future React upgrade. Keeping chai and sinon around after the last suites using them are gone just leaves three testing vocabularies in the tree for new tests to pick from.

The fixture move and the doc rewrite are the tail of the same cleanup: the migration deleted the suites but left their data and their description of the world behind, which is why `src/components/__tests__/` still existed at all.

## How Has This Been Tested?

- Full suite green: `npm test` — 136 suites, 1439 tests passing. Same counts before and after the teardown, i.e. nothing silently stopped running.
- `npm run lint:js` clean; `npx prettier --check` clean on the touched docs.
- The fixture move was not purely mechanical, and the tests caught it: `genreData.ts` had its own relative `../../interfaces` import that broke on the move, failing three suites at compile time until it was repointed. Updating the four importing suites alone would have looked correct.
- Ran the exact `npm run test-jest-file` invocation the rewritten docs advertise, to confirm the documented commands work as written.
- Coverage still reports under the `jest` flag with `coverage/jest/lcov.info` written; the workflow's non-empty check passes.

## Checklist:


tdilauro (approved):
Looks good! 🪣

PR #327, opened 2026-07-16
Drop the write-only customLists state from CustomListsForBook

jonathangreen:
## Description

Removes `state.customLists` from `CustomListsForBook`, along with the `UNSAFE_componentWillReceiveProps` that re-synced it, the `CustomListsForBookState` interface, the constructor's state seed, and the `custom_lists` setting field it fed. Also renames the test that claimed to cover the re-sync to describe what it actually asserts.

Stacked on #325 — please merge that one first. This PR targets `chore/rewrite-container-tests-rtl` and will retarget to `main` automatically once #325 merges.

## Motivation and Context

A review comment on #325 pointed out that the test named "resyncs its lists when the book url changes" was satisfied by prop threading alone and never exercised the lifecycle it targeted. Verifying that turned up the underlying reason: the lifecycle was dead code.

The current-lists column renders from the `customListsForBook` prop — `ProtocolFormField`'s `value` is `this.props.customListsForBook.map(l => l.name)`. Meanwhile `state.customLists` only ever reached `setting.custom_lists`, and nothing reads that field: neither `ProtocolFormField` nor `InputList` mentions it, and the only other occurrences in the tree are the type declarations in `src/interfaces.ts` and two unrelated `mapStateToProps` selectors in `CustomLists.tsx` / `Lanes.tsx` reading a same-named field off Redux state.

That made the state write-only and `UNSAFE_componentWillReceiveProps` a no-op. Deleting it left the entire suite green, which is what confirmed the diagnosis rather than assumed it. Since the re-sync had no observable effect, no test could assert through it — so the honest fix is to remove the state and reframe the test as the prop-threading coverage it always was.

## How Has This Been Tested?

- `npm test` — 1439 tests across 136 suites pass.
- `npx tsc --noEmit` and `npm run lint:js` are clean.
- Mutation check that motivated the change: deleting `UNSAFE_componentWillReceiveProps` on the pre-change code left all 14 `CustomListsForBook` tests passing, including the one named for the re-sync.
- The reframed test now also asserts the previous book's list is *gone* after the rerender, so it fails if prop threading breaks. The assertions use `getByRole("link", ...)` to scope to the current-lists column — a blanket `getByText` match would pick up the add-menu `<option>`, since a list the book is no longer on legitimately becomes addable again.

## Checklist:

### DIFF
--- a/src/components/CustomListsForBook.tsx
+++ b/src/components/CustomListsForBook.tsx
@@ -33,27 +33,18 @@ export interface CustomListsForBookOwnProps {
 }
 
 export interface CustomListsForBookProps
-  extends CustomListsForBookStateProps,
+  extends
+    CustomListsForBookStateProps,
     CustomListsForBookDispatchProps,
     CustomListsForBookOwnProps {}
 
-export interface CustomListsForBookState {
-  customLists?: CustomListData[];
-}
-
 /** Tab on the book details page that shows custom lists a book is on and lets
     an admin add the book to lists or remove the book from lists. */
-export class CustomListsForBook extends React.Component<
-  CustomListsForBookProps,
-  CustomListsForBookState
-> {
+export class CustomListsForBook extends React.Component<CustomListsForBookProps> {
   private addListRef = React.createRef<ProtocolFormField>();
 
   constructor(props: CustomListsForBookProps) {
     super(props);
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
🚀

